PR #3, opened 2016-08-10
Recipe model

tsop14:
- Separates ui and common projects from root.
- Adds model for auto-parsed recipes.

(TeamCity build step set sbt to an old version, which failed to build this branch. This line was removed.)


common/src/main/scala/com/gu/recipeasy/models/Recipe.scala
```diff
@@ -0,0 +1,31 @@
+package com.gu.recipeasy.models
+
+import java.time.OffsetDateTime
+
+case class Recipe(
+  id: String,
```
cb372: title is missing

common/src/main/scala/com/gu/recipeasy/models/Recipe.scala
```diff
@@ -0,0 +1,31 @@
...
+  body: String,
+  serves: Option[Serves],
+  ingredients: Seq[IngredientList],
+  articleId: String,
+  credit: Option[String],
+  publicationDate: OffsetDateTime,
+  status: Status,
+  steps: Option[List[String]]
+)
```
cb372: Also we should add `imageUrl: Option[String]`. We can extend the parser to try to find an image for each recipe, as we discussed.

cb372:
👍  apart from a couple of missing fields

LATaylor-guardian:
👍

PR #4, opened 2016-08-12
Etl

tsop14:
Adds implementation of parsing ingredient lists.
Adds tests.

TODO
-  [ ] have a go at parsing the recipe steps out of the body
- [ ] have a go at finding the image for each recipe in the article
- [ ] rebase against master and integrate with the Recipe model that Emma made today
- [ ] save the parsed recipes to a Dynamo table


etl/src/main/scala/etl/ETL.scala
```diff
@@ -0,0 +1,104 @@
+package etl
```
tsop14: Extract, transform, load

tsop14:
👍

PR #5, opened 2016-08-16
extracts steps and swaps to common recipe model

tsop14:
- Parses recipe method steps if: element is a `<p>` and starts with a digit and is not a `ParaWithListOfShortTexts` (which is a proxy for ingredients).
- Adds test.

- Integrates common recipe models. 

- Ups the character count limit for `ShortListItem` to 120, to correctly parse more ingredients lists.

- Moves etl to root build.sbt file and moves scala test to common dependencies.


etl/src/main/scala/etl/ETL.scala
```diff
@@ -40,7 +43,7 @@ object ETL extends App {
     sys.exit(1)
   }
 
-  val capiKey = args(1)
+  val capiKey = args(0)
```
LATaylor-guardian: How is the script meant to be run now? `sbt run <capi-key>`?

cb372: `sbt etl/run <capi-key>`

LATaylor-guardian: ah!

etl/src/main/scala/etl/ETL.scala
```diff
@@ -79,14 +82,26 @@ object ETL extends App {
...
-      val recipes = rawRecipes.map(RecipeParsing.parseRecipe)
+      val parsedRecipes = rawRecipes.map(RecipeParsing.parseRecipe)
+      val recipes = parsedRecipes.map(r => Recipe(
+        id = r.id,
+        title = r.title,
+        body = r.body,
+        serves = r.serves,
+        ingredientsLists = r.ingredientsLists,
+        articleId = content.id,
+        credit = Some(content.fields.flatMap(_.byline)).getOrElse(None),
```
cb372: Just `credit = content.fields.flatMap(_.byline)` should work fine

etl/src/main/scala/etl/RecipeParsing.scala
```diff
@@ -61,14 +70,20 @@ object RecipeParsing {
...
       } else {
         val ingredients = withoutServingCount.map(text)
-        Some(IngredientList(title = None, ingredients))
+        Some(IngredientsList(title = None, ingredients))
       }
     }
   }
 
+  def guessSteps(body: Seq[Element]): Option[Seq[String]] = {
+    val candidates: Seq[Element] = body.tail.filter(NumberedParagraphs.matches)
```
cb372: Why `body.tail`?

etl/src/main/scala/etl/RecipeParsing.scala
```diff
@@ -117,4 +132,19 @@ object RecipeParsing {
 
   }
 
+  private object NumberedParagraphs {
```
cb372: Could call this `NumberedParagraph` to be consistent with the other matchers. It only matches one paragraph.

etl/src/main/scala/etl/RecipeParsing.scala
tsop14: This needs to be deleted. Previously I wasn't using numbers to discern method stops. `tail` was to excluded the intro text and if there wasn't intro text it would just exclude an ingredient.

cb372:
👍

PR #6, opened 2016-08-17
generates uuid for each recipe

tsop14:
- Replaces placeholder string with id for each recipe.
- Removes comment about images as not attempting to match recipes with images (this can be done manually later).


tsop14:
Don't want a different ID each time, so using a hash. See #7

PR #7, opened 2016-08-17
adds it to recipe using hash of title and publication date

tsop14:
- Gives each recipe a hash of its title and publication date. 
- Removes comment about images as we are no longer going to try to match recipes to images.


LATaylor-guardian:
I think it's important that we still have images for recipes *at some point*. But you're right, this can come later as long as we have a link back to the original article.

cb372:
Having looked at a few recipes, it looks very tricky to automate the process of finding the right image for a recipe. Probably more sensible to show users all the images in the article (with captions) and let them choose the most relevant one (if there is one).

cb372:
👍

PR #8, opened 2016-08-17
makes steps required

tsop14:
- Makes model more consistent by having steps required, as per ingredients lists. If there is no match filtering the body with `NumberedParagraph.matches` will return an empty list.


LATaylor-guardian:
👍

PR #11, opened 2016-08-19
Save recipes to Postgres

cb372:
* Introduce Quill to handle SQL stuff
* Use flyway-sbt for DB migrations
* Add a script to set up DB in dev environment
* Add a README

### DIFF
--- a/etl/src/main/scala/etl/ETL.scala
+++ b/etl/src/main/scala/etl/ETL.scala
@@ -25,6 +25,10 @@ import io.circe.syntax._
 import io.getquill._
 import java.util.Date
 
+import org.postgresql.util.PGobject
+
+import scala.reflect.ClassTag
+
 case class Progress(pagesProcessed: Int, articlesProcessed: Int, recipesFound: Int, articlesWithNoRecipes: List[String]) {
   override def toString: String = s"$pagesProcessed pages processed,\t$articlesProcessed articles processed,\t$recipesFound recipes found,\t${articlesWithNoRecipes.size} articles with no recipes"
 }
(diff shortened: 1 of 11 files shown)


etl/src/main/scala/etl/ETL.scala
```diff
@@ -120,3 +135,35 @@ object ETL extends App {
   }
 
 }
+
+object db {
+
+  lazy val ctx = new JdbcContext[PostgresDialect, SnakeCase]("db.ctx")
+  import ctx._
+
+  val a = quote(query[Recipe].insert)
```
LATaylor-guardian: What is  'a' here? Can we have a more description variable name?

etl/src/main/scala/etl/ETL.scala
```diff
@@ -120,3 +135,35 @@ object ETL extends App {
...
+
+  implicit val encodePublicationDate = mappedEncoding[OffsetDateTime, Date](d => Date.from(d.toInstant))
+  implicit val encodeStatus = mappedEncoding[Status, String](_.toString())
+
+  def jsonbEncoder[T: io.circe.Encoder: ClassTag] =
```
LATaylor-guardian: Public methods should have a return type.

etl/src/main/scala/etl/ETL.scala
```diff
@@ -120,3 +135,35 @@ object ETL extends App {
...
+      pgObj.setType("jsonb")
+      pgObj.setValue(value.asJson.noSpaces)
+      row.setObject(idx, pgObj, Types.OTHER)
+    }, Types.OTHER)
+
+  implicit val servesEncoder: Encoder[Serves] = jsonbEncoder[Serves]
+  implicit val stepsEncoder: Encoder[Steps] = jsonbEncoder[Steps]
+  implicit val ingredientsListsEncoder: Encoder[IngredientsLists] = jsonbEncoder[IngredientsLists]
+
+  def insertAll(recipes: Seq[Recipe]) {
```
LATaylor-guardian: As above.

LATaylor-guardian:
👍

etl/src/main/scala/etl/ETL.scala
tsop14: It's an action. Will change to just that.

PR #12, opened 2016-08-23
moves db to common

tsop14:
- Moves db object from etl to common

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -13,7 +13,8 @@ val commonSettings = Seq(
     "io.circe" %% "circe-generic" % "0.4.1",
     "io.circe" %% "circe-parser" % "0.4.1",
     "org.postgresql" % "postgresql" % "9.4.1208",
-    "io.getquill" %% "quill-jdbc" % "0.9.0"
+    "io.getquill" %% "quill-jdbc" % "0.9.0",
+    "commons-codec" % "commons-codec" % "1.10"
   )
 )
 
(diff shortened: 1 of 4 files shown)


tsop14:
I was not sure where to put / what to call `common/src/main/scala/com/gu/recipeasy/db/actions.scala`??

common/src/main/scala/com/gu/recipeasy/db/actions.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import com.gu.recipeasy.models._
+import io.circe.generic.auto._
+import io.circe.syntax._
+import io.getquill._
+import org.postgresql.util.PGobject
+
+import scala.reflect.ClassTag
+
+object db {
```
cb372: I would call this `object DB` (in uppercase) and rename the file to `DB.scala`.

Once we have more DB tables, we'll probably want to have separate DAOs per table (`Recipes.scala`, `Users.scala`, etc.) but we'll deal with that later.

build.sbt
```diff
@@ -9,11 +9,12 @@ val commonSettings = Seq(
   scalacOptions in doc in Compile := Nil,
   libraryDependencies ++= Seq(
     "org.scalatest" %% "scalatest" % "2.2.6" % Test,
-    "io.circe" %% "circe-core" % "0.4.1",
-    "io.circe" %% "circe-generic" % "0.4.1",
-    "io.circe" %% "circe-parser" % "0.4.1",
+    "io.circe" %% "circe-core" % "0.5.0-M2",
```
tsop14: Upgraded to fix a problem with Cats 😼 (will not work on -M3).

PR #13, opened 2016-08-24
gets recipes from local db and dumps to ui

tsop14:
- Gets the most recent recipe from db and prints its title.
- NB Working with local db.

A most beautiful website! 
screen shot 2016-08-24 at 11 30 43


cb372:
👍

PR #14, opened 2016-08-24
Make DB a class and pass context to it

cb372:
This is so we can use the Play app's conf-magic-generated configuration to build the context.


tsop14:
👍

PR #15, opened 2016-08-26
adds curated model and method to convert raw recipe to curate model

tsop14:
- Creates a `CuratedRecipe` case class.
- Creates a `CuratedRecipeDB` case class, this is the same as the above model except it has [tag ID references](https://github.com/guardian/recipeasy/compare/curated-recipe-model?expand=1#diff-005191d0e10d41249942389adb97dd1dR21) rather than the actual [tags](https://github.com/guardian/recipeasy/compare/curated-recipe-model?expand=1#diff-f294c7a7207f22000831b9fd6cdf8abaR60). 

- Adds a method to convert  raw `Recipe` to a `CuratedRecipe` in preparation for putting into a form on the UI with bootstrap.
- Adds a test

It might be nice to try to do a more betterer conversion, e.g. attempt to separate out the value etc.


common/src/main/scala/com/gu/recipeasy/models/CuratedRecipeDB.scala
```diff
@@ -0,0 +1,22 @@
+package com.gu.recipeasy.models
+
+//this model is different from CuratedRecipe as it holds a list of Tag ids
+//whereas CuratedRecipe has fetches the Tag referenced by the ID from DB
+import java.time.OffsetDateTime
+
+case class EditedRecipeDB(
```
cb372: CuratedRecipeDB?

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
```diff
@@ -0,0 +1,63 @@
...
+case object Teaspoon extends CookingUnit
+case object Tablespoon extends CookingUnit
+case object Pinch extends CookingUnit
+case object Handful extends CookingUnit
+case object Grating extends CookingUnit
+
+case class Tags(list: Seq[Tag])
+
+case class Tag(
+  name: String,
```
cb372: Probably best to standardise on either "name" or "id" for this. I don't really mind which.

cb372:
Looks good!

When we come to convert between `CuratedRecipe` and `CuratedRecipeDB`, my [automagic](https://github.com/cb372/automagic) library might come in handy. I use it to convert between domain and DB models in AMIgo like [this](https://github.com/guardian/amigo/blob/f2326eab47f4c2424a0af073d37422c11aeda5d1/app/models/Bake.scala#L24-L30). Saves a bit of boilerplate.

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipeDB.scala
tsop14: 🙈 will change!

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
tsop14: Are you thinking the ID will be the name? In which case I'll change it to 'name'.

tsop14:
Ah, should I use ✨🔮✨automagic✨🔮✨ for the raw `Recipe` to `CuratedRecipe` too?

ui/test/com/gu/recipeasy/ApplicationSpec.scala
```diff
@@ -0,0 +1,24 @@
+import org.scalatest._
+import collection.mutable.Stack
+import controllers._
+import com.gu.recipeasy.models._
+
+class recipeConversion extends FlatSpec with Matchers {
+  "A raw ingredientis list" should "be converted to a detailed ingredients list" in {
```
LATaylor-guardian: s/ingredientis/ingredients

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipeDB.scala
```diff
@@ -0,0 +1,23 @@
+package com.gu.recipeasy.models
+
+//this model is different from CuratedRecipe as it holds a list of Tag ids
```
LATaylor-guardian: This seperation isn't particularly clear for me coming at it from having not worked on it. Is it possible to make this clearer in the comment, or better yet in the name of the case classes?

ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -6,13 +6,36 @@ import com.gu.recipeasy.views
...
     val newRecipe = db.getNewRecipe
     newRecipe match {
-      case Some(r) => Ok(views.html.app(r))
+      case Some(r) => Ok(views.html.app(recipeTypeConversion.transformRecipe(r)))
       case None => NotFound
     }
   }
 }
+
+object recipeTypeConversion {
```
LATaylor-guardian: Can this live outside of the `Application` controller?

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipeDB.scala
LATaylor-guardian: maybe `RawRecipe` and `CuratedRecipe` as you have said in the tests with a line commenting on what each class represents and in what particular state in the process?

cb372: I tend to nest the DB model inside the domain model's companion object, just so I don't need to think of a good name for it 😉  e.g. https://github.com/guardian/amigo/blob/master/app/models/Bake.scala#L16

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
cb372: OK, let's go with name.

I don't think there's any need for tags to have both a name and a separate ID, if that's what you mean.

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipeDB.scala
LATaylor-guardian: Yeah, this seems like a nice approach to avoid that.

ui/app/com/gu/recipeasy/controllers/Application.scala
tsop14: Should it live in `app/util/`?

cb372:
👍 Sorry this PR dragged on so long!

PR #16, opened 2016-09-07
Adds recipe form and basic ui

tsop14:
screen shot 2016-09-07 at 16 02 47

- updates script to create DB with correct optional / required field settings
- creates a Form model
- creates default cooking unit list
- creates default tags list
- basic form ui
- stylesheet to quickly style nested forms
- js to dynamically add fields

See issue #19 for TODO


tsop14:
FYI the jquery is rewritten in next PR.

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipeDB.scala
```diff
@@ -0,0 +1,23 @@
...
+  body: String,
+  serves: Option[Serves],
+  ingredientsLists: DetailedIngredientsLists,
+  articleId: String,
+  credit: Option[String],
+  publicationDate: OffsetDateTime,
+  status: Status,
+  times: Option[Times],
+  steps: Option[Steps],
+  tags: Option[TagNames]
```
tomrf1: should these be options?

tsop14: No! will change - thanks

ui/public/javascript/editForm.js
```diff
@@ -0,0 +1,144 @@
...
+                $(this).val(quantityGuess[0])
+                $(this).attr("value", quantityGuess[0])
+            }
+        }
+    })
+
+    //guess ingredient unit when field empty
+    $(".ingredients__detail__unit", this).each(function(){
+        var unit = $(this).val()
+        if(unit === "") {
```
tsop14: Note, this wont be working at the moment as the form is setting everything to 'g' by default. TODO add empty string option (this field is an option).

tomrf1:
👍  looking good so far

Issue #19, opened 2016-09-13
First draft TODO


tsop14:
TODO

- [x] create empty steps field when list is empty
- [x] dynamically create ingredient lists/ ingredient
- [x] ignore ingredient.raw and other hidden fields
- [x] add option None for unit
- [x] add author
- [x] add cuisine / meal type
- [x] surface which fields are required
- [x] convert submitted form to DB model
- [x] write to DB
- [x] add CSRF
- [x] Create a DB cleaner that will reset the status of 'pending' recipes after 24hrs.

FORM
- [x] add original recipe page to page
- [x] add pictures
- [x] fill in cuisine, holiday, mealtype selections when present
- [x] make tags multiselect
- [x] add decimals to units
- [x] make raw field editable (if not present this will need to be filled in)
- [x] better labels

PR #20, opened 2016-09-15
finishes form

tsop14:
Finishes the ui form, it now displays all fields and tries to auto fills fields when they are empty.

- nests mappings for tags (to be under play's max mapping limit)
- creates a multiTag template for ui
- simplifies cooking/preparation time fields --> makes everything minutes
- predefines list of cuisines/holidays/meal types (these are all tags)
- adds tags to form
- adds times to form
- adds ability to have empty value for selects (adds empty string)

- guess comment when field empty with js/regex
- guess ingredient when field empty with js/regex

Example of autofilling on form :)
screen shot 2016-09-15 at 15 20 25


tsop14:
ticks several items on #19

shtukas:
:+1: Nice UI :)

PR #21, opened 2016-09-16
turns form back to curated model

tsop14:
Converts form back to a `CuratedRecipe` so ready to write to DB!

- adds noneditable fields to the form as hidden inputs - they still need to be passed through for when the model is converted back.

cleans up code
- makes `CuratedRecipeForm` as similar to `CuratedRecipe` as possible and uses nested mappings to create the form
- uses ✨automagic✨ to convert between `CuratedRecipeForm` and `CuratedRecipe`
- moves Tag related code out to helper object

### DIFF
--- a/ui/app/com/gu/recipeasy/controllers/Application.scala
+++ b/ui/app/com/gu/recipeasy/controllers/Application.scala
@@ -13,6 +13,8 @@ import com.gu.recipeasy.auth.AuthActions
 import com.gu.recipeasy.db._
 import com.gu.recipeasy.models._
 import com.gu.recipeasy.views
+import models._
+import models.CuratedRecipeForm._
 import automagic._
 
 class Application(override val wsClient: WSClient, override val conf: Configuration, db: DB, val messagesApi: MessagesApi) extends Controller with AuthActions with I18nSupport {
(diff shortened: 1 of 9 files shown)


tsop14:
#19  convert submitted form to DB model

ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -163,9 +155,79 @@ object Application {
...
+      Tag.highFibre -> t.highFibre,
+      Tag.nutFree -> t.nutFree,
+      Tag.glutenFree -> t.glutenFree,
+      Tag.dairyFree -> t.dairyFree,
+      Tag.eggFree -> t.eggFree,
+      Tag.vegetarian -> t.vegetarian,
+      Tag.vegan -> t.vegan
+    )
+
+    tags.filter { _._2 == true }.keySet.toSeq
```
tsop14: `.keys` returns a `Set` so would still need a `.toSeq`

ui/app/com/gu/recipeasy/models/CuratedRecipeForm.scala
```diff
@@ -0,0 +1,51 @@
...
+  publicationDate: String,
+  status: String,
+  times: TimesInMins,
+  steps: Seq[String],
+  tags: FormTags
+)
+
+object CuratedRecipeForm {
+
+  def toForm(r: CuratedRecipe): CuratedRecipeForm = {
```
LATaylor-guardian: This should exist on the CuratedRecipe companion object.

tsop14: Yeah, I was trying to keep the UI only model (`CuratedRecipeForm`) out of common. `CuratedRecipe` is in common. Putting `def toForm` in `CuratedRecipe`'s companion object would mean that it had to know about the UI model... 

So, should I move the UI model to common? Or import the ui model to common?

LATaylor-guardian: I thought you said CuratedRecipe was not common? If it is common, then ignore me.

tsop14: I did say I would move it... and then didn't because the DB logic which is in common (as used by ETL) uses `CuratedRecipe`.

LATaylor-guardian:
👍

PR #22, opened 2016-09-20
writes curated recipe to db

tsop14:
- Creates a second DB table, called 'curated_recipe' to store `CuratedRecipe`s. 👯

- Converts submitted forms to a `CuratedRecipe` and then a `CuratedRecipeDB` (which has a simplified version of the tags).

- Writes the `CuratedRecipeDB` to 'curated_recipe' table.

- When a recipe is picked from recipe table to show in UI, its status is set to `Pending`.
- When the recipe is submitted to 'curated_recipe' table, it's corresponding entry in the 'recipe' table has status set to `Curated`. This is to prevent the same recipe being curated twice.


tomrf1:
👍

PR #24, opened 2016-09-21
Remove unnecessary form fields

tsop14:
Trimming the fat! 🏋

- Removes un-editable fields from `CuratedRecipeForm`, `CuratedRecipe`.
- Passes `recipeId` in as a url parameter to keep link between form and original recipe. 
- Moves transformation methods out of application into relevant companion object.

- NB DB migration v3 has been edited. This is NOT a problem because the older version has not been run in PROD. 💃


cb372:
👍 as I watched you do most of this

PR #27, opened 2016-09-21
sorts recipes by newest first

tsop14:
- Updates quill quote to create a psql query that is: ORDER BY r.publication_date **DESC** LIMIT 1


shtukas:
:+1:

PR #30, opened 2016-09-26
adds CSRF

tsop14:
- adds CSRF token

NB [Play migration guide](https://www.playframework.com/documentation/2.5.x/Migration25#CSRF-changes):

> There’s a new config option to bypass the new CSRF protection for requests with certain headers. This config option is turned on by default for the Cookie and Authorization headers, so that REST clients, which typically don’t use session authentication, will still work without having to send a CSRF token.

i.e. cookieless requests are not protected by the CSRF filter.


shtukas:
:+1: We checked together that the CSRF token along with the correct cookie are now required for the request to go through.

PR #32, opened 2016-09-28
Updating README with a note on the install of Postgres

shtukas:
This updates shows how to deal with a failure of `./init-dev-DB.sh` caused by an incorrect install of Postgres.


LATaylor-guardian:
I think we should make sure that we are installing Postgres correctly through the appropriate means rather than adding documentation on how to manually resolve an incorrect install of Postgres.

shtukas:
Cool. As discussed, will modify the install script itself :)

PR #33, opened 2016-09-28
Postgres install update in init-dev-DB.sh

shtukas:
This update ensures that some folders that Postgres needs to run
are actually present before we attemp to start it. This problem
seems to be caused by recent versions of macOS.

The initial set of folders came from
http://stackoverflow.com/questions/25970132/pg-tblspc-missing-after-installation-of-latest-version-of-os-x-yosemite-or-el

and the rest came from actual testing, using a new install of MacOS Sierra.

### DIFF
--- a/init-dev-DB.sh
+++ b/init-dev-DB.sh
@@ -11,6 +11,31 @@ else
   echo Looks like PostgreSQL is already installed.
 fi
 
+# ------------------------------------------------------------------------------
+# The next section repairs possible missing folders whose absence prevent
+# Postgres from starting correctly, This problem occurs with the latest
+# versions of macOS (e.g. Yosemite, El Capitan and Sierra).
+
+DIRS="pg_tblspc
+pg_twophase
+pg_stat
+pg_stat_tmp
+pg_replslot
+pg_snapshots
+pg_stat_tmp
+pg_replslot
+pg_snapshots
+pg_commit_ts
+pg_logical
+pg_logical/snapshots
+pg_logical/mappings"
+for dir in $DIRS
+do
+    mkdir -p "/usr/local/var/postgres/$dir"
+done
+
+# ------------------------------------------------------------------------------
+
 if [ -z "$(pgrep postgres)" ]
 then 
   echo Postgres server does not appear to be running. Starting it as a background process ...


init-dev-DB.sh
```diff
@@ -11,6 +11,65 @@ else
   echo Looks like PostgreSQL is already installed.
 fi
 
+# ------------------------------------------------------------------------------
+# The next section repairs possible missing folders whose absence prevent
+# Postgres from starting correctly, This problem occurs with the latest
+# versions of macOS (e.g. Yosemite, El Capitan and Sierra).
+
+if [ ! -d "/usr/local/var/postgres/pg_tblspc" ]; then
```
LATaylor-guardian: Maybe this can be done a little less verbosely by iterating over a list (of files) and using that to check if it exists and if not creating it?

shtukas: Ok!

init-dev-DB.sh
```diff
@@ -11,6 +11,65 @@ else
...
+fi
+
+if [ ! -d "/usr/local/var/postgres/pg_logical/snapshots" ]; then
+    mkdir /usr/local/var/postgres/pg_logical/snapshots
+fi
+
+if [ ! -d "/usr/local/var/postgres/pg_logical/mappings" ]; then
+    mkdir /usr/local/var/postgres/pg_logical/mappings
+fi
+
```
cb372:
```
for dir in pg_tblspc pg_twophase pg_stat pg_stat_tmp pg_replslot pg_snapshots pg_commit_ts pg_logical/snapshots pg_logical/mappings; do
  mkdir -p "/usr/local/var/postgres/$dir"
done
```

#justsayin

shtukas: Happy with the logic of the loop but I think we still need to perform a file check because some of those folders might actually already exist [1] (and `mkdir` complains when we attempt to create an existing folder).

[1] Some of then already existed for me

oilnam: That's why Chris suggested to use `mkdir -p`

shtukas: Thanks! My mistake!

LATaylor-guardian:
If you've tested it works as expected 👍

shtukas:
Yes I did. Thanks!

PR #37, opened 2016-09-29
creates an image table

tsop14:
- crates an `Image` case class with: mediaId, articleId and assetUrl
- creates image table so we can display all images from an article on the page with a recipe from that article for user to select
- edits ETL to get element fields and write the images to image table

### DIFF
--- a/common/src/main/resources/db/migration/V4__create_image_table.sql
+++ b/common/src/main/resources/db/migration/V4__create_image_table.sql
@@ -1,0 +1,8 @@
+CREATE TABLE image (
+    -- cannot use media_id as primary key as some images are used twice --
+    id serial primary key,
+    media_id varchar(80) not null,
+    article_id text not null,
+    asset_url text not null
+);
+
--- a/common/src/main/resources/db/migration/V4__create_image_table.sql
+++ b/common/src/main/resources/db/migration/V4__create_image_table.sql
@@ -3,6 +3,7 @@ CREATE TABLE image (
     id serial primary key,
     media_id varchar(80) not null,
     article_id text not null,
-    asset_url text not null
+    asset_url text not null,
+    alt_text text
 );
 
--- a/common/src/main/scala/com/gu/recipeasy/db/DB.scala
+++ b/common/src/main/scala/com/gu/recipeasy/db/DB.scala
@@ -50,17 +50,18 @@ class DB(ctx: JdbcContext[PostgresDialect, SnakeCase]) {
   }
 
   private implicit val servesEncoder: Encoder[Serves] = jsonbEncoder[Serves]
-  private implicit val stepsEncoder: Encoder[Steps] = jsonbEncoder[Steps]
   private implicit val ingredientsListsEncoder: Encoder[IngredientsLists] = jsonbEncoder[IngredientsLists]
   private implicit val detailedIngredientsListsEncoder: Encoder[DetailedIngredientsLists] = jsonbEncoder[DetailedIngredientsLists]
   private implicit val timesEncoder: Encoder[TimesInMins] = jsonbEncoder[TimesInMins]
+  private implicit val stepsEncoder: Encoder[Steps] = jsonbEncoder[Steps]
   private implicit val tagsEncoder: Encoder[TagNames] = jsonbEncoder[TagNames]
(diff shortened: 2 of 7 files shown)


common/src/main/resources/db/migration/V4__create_image_table.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE image (
+    -- cannot use media_id as primary key as some images are used twice --
```
LATaylor-guardian: Sensible.

common/src/main/scala/com/gu/recipeasy/db/DB.scala
```diff
@@ -74,6 +75,18 @@ class DB(ctx: JdbcContext[PostgresDialect, SnakeCase]) {
...
   }
 
+  def insertImages(images: List[Image]): Unit = {
+    try {
+      val action = quote {
+        liftQuery(images).foreach(i => query[Image].insert(i))
+      }
+      ctx.run(action)
+    } catch {
+      case e: java.sql.BatchUpdateException => throw e.getNextException
```
LATaylor-guardian: In the event of an exception we will want to log it and something useful.

common/src/main/scala/com/gu/recipeasy/db/DB.scala
```diff
@@ -74,6 +75,18 @@ class DB(ctx: JdbcContext[PostgresDialect, SnakeCase]) {
     }
   }
 
+  def insertImages(images: List[Image]): Unit = {
```
LATaylor-guardian: Is there anything other than logging we should be doing if this fails?

LATaylor-guardian: Can this take a Seq[Image] so you don't have to do .toList further down when calling the function?

etl/src/main/scala/etl/ImageExtraction.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  def getImages(content: Content, articleId: String): Seq[Image] = {
+    val images = findImageElements(content)
+    images match {
+      case Some(images) => images.flatMap(i => extractImageData(i, articleId))
+      case None => Nil
+    }
+  }
+
+  private def findImageElements(content: Content): Option[Seq[Element]] = {
```
LATaylor-guardian: This can return a Seq[Element] as opposed to an Option[Seq[Element]] and return Nil if it doesn't find any imageElements given you are matching on None above and returning Nil.

etl/src/main/scala/etl/ImageExtraction.scala
```diff
@@ -0,0 +1,42 @@
...
+    } yield new Image(
+      image.id,
+      articleId,
+      assetUrl,
+      altText
+    )
+  }
+
+  private def findLargestAsset(element: Element): Option[Asset] = {
+    element.assets.sortBy(_.typeData.flatMap(_.width)).lastOption
```
LATaylor-guardian: Nice :)

common/src/main/scala/com/gu/recipeasy/db/DB.scala
tsop14: This is a one off process that is run, not sure of best logging practice? At the moment it goes as far as `println`s.

common/src/main/scala/com/gu/recipeasy/db/DB.scala
tsop14: It would fail to post all the images but only for one article at a time. Might be worth tracking the number of articles that fail?

tsop14: FYI we have recipes from 6528 articles and images from 5097 articles.

tsop14: Quill says no (I think).

etl/src/main/scala/etl/ImageExtraction.scala
tsop14: Done!

alastairjardine:
Is there any UI to review, or am I jumping this particular gun?

LATaylor-guardian:
👍

tsop14:
@alastairjardine no ui in this one.

PR #38, opened 2016-09-29
New route to see a recipe by its identifier

shtukas:


tsop14:
What happens if two people open the same recipe? The first person will not realise all their changes will be overwritten. Do we want to handle this?

LATaylor-guardian:
@shtukas If you can view the recipe like this are you able to edit it also? Also, whats the reason for adding this endpoint - can you update the PR description with it please.

shtukas:
@tsop14 
So... as discussed, this is currently a private route of us (admins) to access a recipe we have the id of. Not meant to be a part of the user curation workflow.

shtukas:
> What happens if two people open the same recipe?
As discussed, let me remove the edit button [Submit Recipe] on that page (and the [Skip this recipe] button as well).

shtukas:
Done. The buttons now only show up on the standard curation page, and not on the new "view only" page :)

tsop14:
This could be useful to view articles that are consistently skipped?

👍

ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -29,7 +29,7 @@ class Application(override val wsClient: WSClient, override val conf: Configurat
         val curatedRecipe = CuratedRecipe.fromRecipe(r)
         val curatedRecipeForm = CuratedRecipeForm.toForm(curatedRecipe)
         db.setRecipeStatus(r.id, "Pending")
-        Ok(views.html.recipe(createCuratedRecipeForm.fill(curatedRecipeForm), r.id, r.body, r.articleId))
+        Ok(views.html.recipe(createCuratedRecipeForm.fill(curatedRecipeForm), r.id, r.body, r.articleId, true))
```
LATaylor-guardian: For readability this would be better written as: 
Ok(views.html.recipe(createCuratedRecipeForm.fill(curatedRecipeForm), r.id, r.body, r.articleId, shouldShowButtons = true))

ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -54,6 +54,20 @@ class Application(override val wsClient: WSClient, override val conf: Configurat
...
   }
+
+  def viewRecipe(recipeId: String) = Action { implicit request =>
+    val newRecipe = db.getRecipe(recipeId)
+    newRecipe match {
+      case Some(r) => {
+        val curatedRecipe = CuratedRecipe.fromRecipe(r)
+        val curatedRecipeForm = CuratedRecipeForm.toForm(curatedRecipe)
+        db.setRecipeStatus(r.id, "Pending")
+        Ok(views.html.recipe(createCuratedRecipeForm.fill(curatedRecipeForm), r.id, r.body, r.articleId, false))
```
LATaylor-guardian: As above.

ui/app/com/gu/recipeasy/controllers/Application.scala
shtukas: Cool, will send a separate PR.

PR #42, opened 2016-10-03
adds image to ui and form

tsop14:
- adds images to curated recipe model
- updates `CuratedRecipe` DB table to include image field (adds encoder to add images)

- adds function to fetch images associated with an article from table
- adds images from parent article for each recipe shown on ui

- creates hidden and custom multiImage bootstrap template for adding to form

-creates button to add image to the form

@alastairjardine UI in next PR 

addimage


tomrf1:
👍  Nice!

PR #43, opened 2016-10-04
updates etl to include adding images to table

tsop14:
- collects list of articleIds already in the db
- filters content against this list
- if successfully parse recipes from article, add images from that article to db

The DB in prod has been updated. 

JS correction to clear field of dynamically added ingredient.

### DIFF
--- a/etl/src/main/scala/etl/ETL.scala
+++ b/etl/src/main/scala/etl/ETL.scala
@@ -89,13 +89,16 @@ object ETL extends App {
   }
 
   def processPage(contents: List[Content])(implicit db: DB): Progress = {
-    val progress = contents.foldMap { content =>
+    val existingArticlesIds = db.existingArticlesIds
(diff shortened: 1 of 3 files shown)


etl/src/main/scala/etl/ETL.scala
```diff
@@ -89,13 +89,16 @@ object ETL extends App {
   }
 
   def processPage(contents: List[Content])(implicit db: DB): Progress = {
-    val progress = contents.foldMap { content =>
+    val existingArticlesIds = db.existingArticlesIds
```
LATaylor-guardian: This should be `db.existingArticlesIds()` as it has a side effect of calling the DB.

LATaylor-guardian:
👍

etl/src/main/scala/etl/ETL.scala
tsop14: Thanks! done

PR #44, opened 2016-10-04
fixes broken image form

tsop14:
Fixes broken form submission - you cannot submit an image as part of form. 

- displays an image which is not part of the form
- creates a hidden input with the assetUrl to submit


LATaylor-guardian:
👍

PR #50, opened 2016-10-06
Styling

tsop14:
This PR makes styling changes as per AJ's design and writes logic for using keyboard shortcuts to add ingredients/ steps. 

- upgrades to bootstrap 4
- renames directory 'bootstrap3' to 'helper', which had made all helper html templates look like new code
- replaces glyphicon (not supported in bootstrap 4) with font-awesome

- corrects hard coded html in editForm.js so buttons to add/remove fields work with styling changes
- uses mousetrap to add keyboard short cuts
- uses jquery selection to add highlighted text to form

Current keyboard shortcuts
'i' adds a single ingredient to end of first list and tries to parse
'I' - adds a list of ingredients to last list and tries to parse
's' - adds a step

KNOWN BUGS
This will be fixed but don't want to block PR
- keyboard shortcuts do not remove empty fields
- 'i' and 'I' add to different lists
- styling broken on tag select

INGREDIENTS
ingredient
- the auto parsed ingredients
- adding blank ingredient and ingredient list
- removing ingredient and ingredient list
- adding a list of ingredients (keyboard shortcut I)
- adding an ingredient (keyboard shortcut i)

STEPS
steps
- adding / removing steps
- adding step (keyboard shortcut s)

IMAGES
add-images
- adding / removing images
- NB my LOCAL `image` DB table has duplicate images so in videos images are duplicated, this should not happen in PROD

SOME MODERN ART I MADE WHILE TRYING TO RECORD MY SCREEN
ingredients
- wow
- poignant


alastairjardine:
👍  for the modern art ;)

alastairjardine:
Rest of the PR looks great! Anywhere I can have a play?

mchv:
💯 for the art

mchv:
Difficult to review but nothing to worry for me 👍 
Screenshots looks awesome!

alastairjardine:
Few thoughts:
On keyboard shortcuts:
- Is the add as list a uppercase `L` or an uppercase `I`? It's confusing!
- Switch the step shortcut to m, as it becomes mnemonic with method
- how about: `l` for add as list; `m` for add to method; `i` for ingredients. They're also close on the keyboard (but on the right hand, rather than left, which would be more usable 🤔)

Other
- I think we could be clearer with the "Add to recipe" button, perhaps "Set at recipe picture"?
- align the 'edit' for the original ingredient left, so it's just after the ingredient
- can you update the dropdown so it reads grams (g), tablespoons (tbsp) etc, or is that for another PR?
- let's have a play around with the sizing of various elements
- could we add in the tiny field headings?
screen shot 2016-10-07 at 10 18 19

PR #52, opened 2016-10-07
Riff raff

tsop14:
Upgrades to latest version of sbt-riffraff-artifact
Adds `riffRaffManifestVcsUrl` as TeamCity checks out on the server then copies the directory over without git info and so need to specify the value.


tomrf1:
👍

LATaylor-guardian:
👍 🔍

Issue #46, opened 2016-10-06
Makes/ serves field


tsop14:
- [x] update model to include string of "makes" or "serves"
- [x] write db migration for `curated-recipe` table to add in field
- [x] add to form

tsop14:
![servings](https://cloud.githubusercontent.com/assets/8484757/19158438/9d1e5b80-8be0-11e6-8e81-3c51af7a1b3c.png)

PR #59, opened 2016-10-12
adds serves/makes field

tsop14:
**MODEL**
- creates `detailedServes` class which includes a field with portionType
- create sealed trait portion type which is either 'serves' or 'makes'
- uses circe to create JSON for adding to curated_recipe table:
```
{"portion": "ServesType", "quantity": {"to": 2, "from": 2}}
```

**UI**
- if the user checks 'No serving information', the serves fields are disabed. This preserves their values on the UI but will not post with form
- if the user selects either radio button AND leaves quantites empty, they cannot submit the form
- if the user does not check 'no serving information' AND has not selected a radio button they can still submit the form

required

TODO
- [ ] clean prod database; 36 recipes have been added to DB with old `Serves` model.

### DIFF
--- a/common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
+++ b/common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
@@ -26,11 +26,8 @@ case class DetailedServes(
 )
 
 object DetailedServes {
-  def fromServes(s: Option[Serves]): Option[DetailedServes] = {
-    s match {
-      case Some(s) => Some(DetailedServes(ServesType, s))
-      case None => None
-    }
+  def fromServes(serves: Option[Serves]): Option[DetailedServes] = {
+    serves.map(s => DetailedServes(ServesType, s))
   }
 }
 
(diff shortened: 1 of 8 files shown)


tsop14:
#46

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
```diff
@@ -20,6 +20,42 @@ case class CuratedRecipe(
...
 )
 
+case class DetailedServes(
+  portion: PortionType,
+  quantity: Serves
+)
+
+object DetailedServes {
+  def fromServes(s: Option[Serves]): Option[DetailedServes] = {
+    s match {
```
LATaylor-guardian: `s` is already an Option so you can map on it like so:

```
serves.map(s=> DetailedServes(ServesType, Some(s)))
```

tsop14:
screen shot 2016-10-12 at 16 37 14

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
tsop14: Done.

LATaylor-guardian:
👍

PR #60, opened 2016-10-12
Fix continous integration

mchv:
- you could now deploy with `riff-raff` with the correct `branch` version and with the correct name `Off-platform::recipeasy`
- Update content-api to`10.5`
- Update play framework to `2.5.x`


LATaylor-guardian:
👍

PR #62, opened 2016-10-13
Shareable url

mchv:
When accessing `/recipe/curate` you will be redirected to `/recipe/curate/<recipedId>`, which means, you can share or bookmark `url` of a recipe to edit it later or show it to someone.

### DIFF
--- a/ui/app/com/gu/recipeasy/controllers/Application.scala
+++ b/ui/app/com/gu/recipeasy/controllers/Application.scala
@@ -21,22 +21,45 @@ class Application(override val wsClient: WSClient, override val conf: Configurat
     Ok(views.html.app("Recipeasy"))
   }
 
-  def curateRecipePage = Action { implicit request =>
+  def curateNewRecipe = Action { implicit request =>
     val newRecipe = db.getNewRecipe
     newRecipe match {
+      case Some(r) => Redirect(routes.Application.curateRecipe(r.id))
+      case None => NotFound
+    }
+  }
+
+  def curateRecipe(id: String) = Action { implicit request =>
+    val recipe = db.getRecipe(id)
+    view(recipe, editable = true)
+  }
+
+  def viewRecipe(id: String) = Action { implicit request =>
+    val recipe = db.getRecipe(id)
+    view(recipe, editable = false)
+  }
+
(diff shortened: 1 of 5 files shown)


ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -22,21 +22,38 @@ class Application(override val wsClient: WSClient, override val conf: Configurat
...
+    }
+  }
+
+  def curateRecipe(id: String) = Action { implicit request =>
+    val recipe = db.getRecipe(id)
+    view(recipe, editable = true)
+  }
+
+  def viewRecipe(id: String) = Action { implicit request =>
+    val recipe = db.getRecipe(id)
```
tsop14: This will fetch the recipe from the `recipe` table. Not the `curated_recipe` table. When a user edits a recipe and submits it their changes are saved to `curated_recipe` table, this means if they send the url to someone else, the recipe which gets returned will not have their edits.

mchv: @tsop14 thanks need to think about that. I will update the code

mchv: @tsop14 the code only retrieve am `id` there which is then use top retrieve the `curated_recipe` in `view`:
```scala
val curatedRecipe = CuratedRecipe.fromRecipe(r)
```
So I think this code will retrieve all the saved content of the `curated_recipe` table

tsop14: Need to write this method: https://github.com/guardian/recipeasy/blob/master/common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala#L181 and add a method to extract from `curated_recipe` table.

mchv: @tsop14 Cheers. It perfectly make sense.

tsop14: This has been added now #68

mchv: Just updated the code around this

LATaylor-guardian:
👍

Issue #49, opened 2016-10-06
Units


tsop14:
At the moment units are displayed as abbreviations e.g. 'tbsp'

List units in full e.g. 'tablespoon (tbsp)'

- [x] change select list
- [x] update guessUnit method in editForm.js

PR #63, opened 2016-10-13
adds more units, orders in ui

tsop14:
- creates more units
- creates display name for units
- creates subsections to order selection on UI

- fixes styling on videos in original recipe body

screen shot 2016-10-13 at 15 58 53

### DIFF
--- a/common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
+++ b/common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
@@ -85,26 +85,46 @@ case class TimesInMins(
 
 sealed trait CookingUnit {
   def abbreviation: String
+  def displayName: String
 }
 
 object CookingUnit {
   val unitMap = Map(
-    "g" -> Gram,
-    "ml" -> Milliltre,
-    "l" -> Litre,
-    "oz" -> Ounces,
-    "floz" -> FluidOunces,
     "cup" -> Cup,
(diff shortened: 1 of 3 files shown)


tsop14:
#49

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
```diff
@@ -85,27 +85,47 @@ case class TimesInMins(
 
 sealed trait CookingUnit {
   def abbreviation: String
+  def displayName: String
 }
 
 object CookingUnit {
   val unitMap = Map(
+    "cup" -> Cup,
```
LATaylor-guardian: The key in this map corresponds to the abbreviation on the corresponding object, correct? If so, we should reference Cup.abbreviation as the key here so we don't have it in two separate places.

alastairjardine:
`FlOz` to `Fl oz`

Other than that 👍

alastairjardine:
Oh Oh!

We need to add dessert spoon!

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
tsop14: ⚡️ V. sensible! Changed

tsop14:
@alastairjardine dessert spoon added.

LATaylor-guardian:
Fine from a code point of view 👍

PR #67, opened 2016-10-17
Move recipeasy.service file to ui project.

LATaylor-guardian:
The sbt-riff-raff-artifact plugin was unable to pickup the recipeasy.service file as it was in the root directory and not in the ui directory where `sbt riffRaffUpload` was being run from.


tsop14:
👍 thanks!

mchv:
💯

PR #68, opened 2016-10-17
writes logic to get curated recipe from curated_recipe table

tsop14:
- writes quill query to get recipe from db
- writes transformation from `curatedRecipeDB` to `curatedRecipe`, which needs tags to be expanded from `TagNames` to `Tags` e.g. 

`Seq("Christmas")` to `Seq(Tag("Christmas", "holiday")`

### DIFF
--- a/common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
+++ b/common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
@@ -230,9 +230,9 @@ object CuratedRecipe {
   def getFullTags(tags: TagNames): Tags = {
     Tags(tags.list.collect {
       case t if Tag.cuisines.contains(t) => Tag(t, "cuisines")
-      case t if Tag.cuisines.contains(t) => Tag(t, "category")
+      case t if Tag.category.contains(t) => Tag(t, "category")
       case t if Tag.holidays.contains(t) => Tag(t, "holidays")
-      case t if Tag.cuisines.contains(t) => Tag(t, "dietary")
+      case t if Tag.dietary.contains(t) => Tag(t, "dietary")
     })
   }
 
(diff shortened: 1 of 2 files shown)


common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
```diff
@@ -217,12 +217,24 @@ object CuratedRecipe {
...
   def getTagNames(tags: Tags): TagNames = {
     TagNames(tags.list.map(t => t.name))
   }
 
+  def getFullTags(tags: TagNames): Tags = {
+    Tags(tags.list.collect {
+      case t if Tag.cuisines.contains(t) => Tag(t, "cuisines")
+      case t if Tag.cuisines.contains(t) => Tag(t, "category")
+      case t if Tag.holidays.contains(t) => Tag(t, "holidays")
+      case t if Tag.cuisines.contains(t) => Tag(t, "dietary")
```
tsop14: Should this have a `case _ => None` ?

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
```diff
@@ -217,12 +217,24 @@ object CuratedRecipe {
...
   }
 
   def getTagNames(tags: Tags): TagNames = {
     TagNames(tags.list.map(t => t.name))
   }
 
+  def getFullTags(tags: TagNames): Tags = {
+    Tags(tags.list.collect {
+      case t if Tag.cuisines.contains(t) => Tag(t, "cuisines")
+      case t if Tag.cuisines.contains(t) => Tag(t, "category")
```
LATaylor-guardian: Should this be `Tag.category`?

LATaylor-guardian:
👍

PR #69, opened 2016-10-17
convers app.json cloud formation to app.yml

tsop14:
- updates cf for app to yml file

Updated in PROD and health checked new instance - looks fine

@LATaylor-guardian


LATaylor-guardian:
👍

PR #70, opened 2016-10-17
converts rds.json cloudformation to yml

tsop14:
- Converts rds.json to rds.yml 

This has not been tested in PROD as wanted some 👀 first


LATaylor-guardian:
Looks okay to me.

PR #71, opened 2016-10-18
adds logging to ELK stack

tsop14:
Sets up logging. Currently, recipeasy logs the PSQL commands create by quill and when a curated recipe is loaded. Tested  locally ☺︎

- sends logs to ELK stack

TODO 
- [ ] update CF


LATaylor-guardian:
Looks okay to me, but the proof is in the pudding 👍  Don't forget to update the config on PROD :)

tsop14:
Config updated! ✅

PR #75, opened 2016-10-18
adds byline field to capi request

tsop14:
If we run ETL again, this will populate the credit field. 

- adds byline field to capi request


shtukas:
:+1:

PR #76, opened 2016-10-19
adds ability to update existing articles if they are not already edited

tsop14:
Adds ability to safely run ETL to add new articles, and update existing recipes if it is safe to do so.

An article is safe to update if none of its corresponding recipes have been edited (i.e. status field is 'New' and not 'Pending' or 'Curated' or 'Impossible'). This is because once an article has been edited future edits will be built upon this version and not the original version parsed by the ETL. The new ETL could effect the number of recipes parsed from an article and update the wrong recipe in the table.

List of safe to update recipes is found by making a diff of list of article_ids with existing recipes and list of article_ids for recipes with status != 'New'. 

- splits up process to allow for updating and inserting
- amends ETL to update safe existing articles
- images are only added if the article is fresh content

Tested locally. 

Rough diagram of what the ETL does for reference: 

etl 1

### DIFF
--- a/etl/src/main/scala/etl/ETL.scala
+++ b/etl/src/main/scala/etl/ETL.scala
@@ -89,9 +89,7 @@ object ETL extends App {
   }
 
   def processPage(contents: List[Content])(implicit db: DB): Progress = {
-
-    processPageWhenInserting(contents)
-    processPageWhenUpdating(contents)
+    Progress.progressMonoid.combine(processPageWhenInserting(contents), processPageWhenUpdating(contents))
   }
 
   def processPageWhenInserting(contents: List[Content])(implicit db: DB): Progress = {
--- a/etl/src/main/scala/etl/ETL.scala
+++ b/etl/src/main/scala/etl/ETL.scala
@@ -93,8 +93,7 @@ object ETL extends App {
   }
 
   def processPageWhenInserting(contents: List[Content])(implicit db: DB): Progress = {
-    val existingArticlesIds = db.existingArticlesIds()
-    val freshContent = contents.filterNot(c => existingArticlesIds.contains(c.id))
+    val freshContent = contents.filterNot(c => db.existingArticlesIds contains(c.id))
 
     val progress = freshContent.foldMap { content =>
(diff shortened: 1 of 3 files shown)


oilnam:
👍  and 💯 for the great diagram!

etl/src/main/scala/etl/ETL.scala
```diff
@@ -89,50 +89,84 @@ object ETL extends App {
...
+      makeProgress(recipes, content)
+
     }
     progress.copy(pagesProcessed = 1)
   }
 
+  def getRecipes(content: Content)(insertImages: List[ImageDB] => Unit): Seq[Recipe] = {
+
+    val rawRecipes = RecipeExtraction.findRecipes(content.webTitle, content.fields.flatMap(_.body).getOrElse(""))
+    if (rawRecipes.nonEmpty) {
```
oilnam: If you get an `Option` back you can map on it instead of doing a `getOrElse` followed by a `nonEmpty`

etl/src/main/scala/etl/ETL.scala
```diff
@@ -89,50 +89,84 @@ object ETL extends App {
...
+      val images = ImageExtraction.getImages(content, content.id).toList
+      insertImages(images)
+    }
+    val parsedRecipes = rawRecipes.map(RecipeParsing.parseRecipe)
+    val publicationDate = content.webPublicationDate.map(time => OffsetDateTime.parse(time.iso8601)).getOrElse(OffsetDateTime.now)
+
+    parsedRecipes.zipWithIndex.map {
+      case (r, i) =>
+        // include index because some recipes inside the same article have duplicate titles! (due to a parser bug)
+        val id = md5Hex(s"${r.title} :: ${content.id} :: $i")
```
oilnam: you're only using this one so no need to have it on a `val` I guess

tsop14: I want to keep the definition out of the instance creation because I think it is easier to understand.

tsop14:
@oilnam updated as per your recommendations!

PR #77, opened 2016-10-20
removes empty ingredients list when adding new one w keyboard shortcut

tsop14:
- checks if ingredient list being appended to is emtpy, if yes removes after adding new list
- makes display value "Gram (g)" singular - to be consistent with other units
- adds value to aws-region in conf to prevent `IllegalArgumentException: Cannot create enum from  value!` error

NOW ☺︎
removes empty list

BEFORE ☠
removes empty list fail


LATaylor-guardian:
👍

PR #78, opened 2016-10-20
only inserts images once per article

tsop14:
If an article has recipes extract, we only want to add images from the article to `image` table once.

- only runs add images once per piece of content rather than for each recipe
- prod DB has been updated to remove duplicated images.


oilnam:
Ooooh I didn't realize you wanted to do that once-a-time rather than for every recipe :) 

Sorry, my fault.

👍

PR #79, opened 2016-10-21
we should not use history but the link to curate a new recipe

mchv:


mchv:
I am going to merge this for being ready at `3pm`

LATaylor-guardian:
👍

PR #80, opened 2016-10-24
Fixes bugs - as per feedback

tsop14:
ETL
- Removes numbers from steps using regex
- Updates test
- [ ] Update PROD

UI
- put images before tags on RHS of form to increase visibility
- fixes broken autosuggest on chef/cook field
- autoheight on steps, (this includes a listener which is very noisy so have debounced key events)

screen shot 2016-10-24 at 18 31 50

### DIFF
--- a/etl/src/main/scala/etl/RecipeParsing.scala
+++ b/etl/src/main/scala/etl/RecipeParsing.scala
@@ -78,12 +78,12 @@ object RecipeParsing {
     }
   }
 
+  val stepPattern = new Regex("""(^\d+.?\s?)(.*+)""", "number", "text")
+
   def guessSteps(body: Seq[Element]): Seq[String] = {
     val candidates: Seq[Element] = body.filter(NumberedParagraph.matches)
     val stepsFound = candidates.map(_.text).filter(_.nonEmpty)
-    val pattern = new scala.util.matching.Regex("""(^\d+.?\s?)(.*+)""", "number", "step")
-
-    stepsFound.map(s => pattern replaceAllIn (s, m => m.group("step")))
+    stepsFound.map(s => stepPattern replaceAllIn (s, m => m.group("text")))
   }
 
   /**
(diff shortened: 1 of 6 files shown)


etl/src/main/scala/etl/RecipeParsing.scala
```diff
@@ -79,8 +80,10 @@ object RecipeParsing {
 
   def guessSteps(body: Seq[Element]): Seq[String] = {
     val candidates: Seq[Element] = body.filter(NumberedParagraph.matches)
-    candidates.map(_.text).filter(_.nonEmpty)
+    val stepsFound = candidates.map(_.text).filter(_.nonEmpty)
+    val pattern = new scala.util.matching.Regex("""(^\d+.?\s?)(.*+)""", "number", "step")
```
LATaylor-guardian: Do you need the full package name here given it's imported above?

LATaylor-guardian: Also, the regular expression only needs to be created once so can be pulled out of the function and placed on the object to ensure this. This is good practice as it can be expensive to compile regular expressions (in Java at least as they are compiled into an internal data structure) so this way we only incur the cost once rather than every time the function is called.

LATaylor-guardian:
Other than my comments 👍

etl/src/main/scala/etl/RecipeParsing.scala
mchv: Put the regex in an object as its compilaton is perf costly, see my last ophan PR

tsop14: Thanks! Will do.

tsop14: Yep all true

> Scala regex class delegates to the java.util.regex package of the Java Platform.

> An instance of Regex represents a compiled regular expression pattern. Since compilation is expensive, frequently used Regex es should be constructed once, outside of loops and perhaps in a companion object.

PR #82, opened 2016-10-28
Add support for date comparison

tsop14:
Makes it possible to compare recipe publication date using quill by adding support for < and > operators. 
- Db context has been abstracted into a wrapper which defines the trait quote. 
- Context created then extends this trait. 
- Updates to UI and ETL to support change.

Thanks @LATaylor-guardian @tomrf1

### DIFF
--- a/common/src/main/scala/com/gu/recipeasy/db/ContextWrapper.scala
+++ b/common/src/main/scala/com/gu/recipeasy/db/ContextWrapper.scala
@@ -1,0 +1,18 @@
+package com.gu.recipeas.db
+
+import com.typesafe.config.{ Config => TypesafeConfig }
+import io.getquill.{ ImplicitQuery, JdbcContext, PostgresDialect, SnakeCase }
+import java.time.OffsetDateTime
+
+trait ContextWrapper {
+  val config: TypesafeConfig
+  lazy val dbContext = new JdbcContext[PostgresDialect, SnakeCase](config.getConfig("db.ctx")) with ImplicitQuery with Quotes
+
+  trait Quotes {
+    this: JdbcContext[_, _] =>
+    implicit class DateTimeQuotes(left: OffsetDateTime) {
+      def >(right: OffsetDateTime) = quote(infix"$left > $right".as[Boolean])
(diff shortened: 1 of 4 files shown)


etl/src/main/scala/etl/ETL.scala
```diff
@@ -53,9 +54,12 @@ object ETL extends App {
     .showFields("main,body,byline")
     .showElements("image")
 
-  val dbContext = new JdbcContext[PostgresDialect, SnakeCase]("db.ctx")
+  val contextWrapper = new ContextWrapper { val config = ConfigFactory.load() }
+  val test = ConfigFactory.load()
```
tomrf1: debugging line?

common/src/main/scala/com/gu/recipeasy/db/ContextWrapper.scala
```diff
@@ -0,0 +1,19 @@
...
+import io.getquill.{ ImplicitQuery, JdbcContext, PostgresDialect, SnakeCase }
+import java.time.OffsetDateTime
+
+trait ContextWrapper {
+  val config: TypesafeConfig
+  lazy val dbContext = new JdbcContext[PostgresDialect, SnakeCase](config.getConfig("db.ctx")) with ImplicitQuery with Quotes
+
+  trait Quotes {
+    this: JdbcContext[_, _] =>
+    implicit class DateTimeQuotes(left: OffsetDateTime) {
```
mchv: need to extend `AnyVal`

mchv:
Very nice workaround 💯

etl/src/main/scala/etl/ETL.scala
tsop14: 👀 Quite right! Removing

common/src/main/scala/com/gu/recipeasy/db/ContextWrapper.scala
tsop14: Done

tomrf1:
👍

tsop14:
👍

PR #83, opened 2016-11-03
Correct bug in ingredient list management

shtukas:
This corrects a bug by which the values of input elements of an ingredient list form where not correctly retrieved. The incorrect code used `$(this).find("input").val()` instead of the correct `$(this).val()`. Also note that the CSS class used to select the elements of the form has been corrected. The incorrect code used `ingredient` instead of the correct `ingredient__detail`, and the logic of the function has been reworked to make it easier to understand.


alastairjardine:
great, thanks Pascal. Good spot. Can't comment on the code, though!

LATaylor-guardian:
:+1

PR #88, opened 2016-11-08
hours+minutes update in recipe times

shtukas:
From the User Interface, this commit moves from

screen shot 2016-11-08 at 17 44 20

to

screen shot 2016-11-08 at 17 46 51

Note that this was done without changing the database model (essentially without changing the definition of `case class CuratedRecipeDB`)

Note that this is the first of two commits. The second commit will ensure that minutes always remain in the range `[0,60]`.

### DIFF
--- a/common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
+++ b/common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
@@ -6,19 +6,6 @@ import cats.data.Xor
 import CuratedRecipeDB._
 import ImageDB._
 
-case class CuratedRecipe(
-  id: Long,
-  recipeId: String,
-  title: String,
-  serves: Option[DetailedServes],
-  ingredientsLists: DetailedIngredientsLists,
-  credit: Option[String],
-  times: TimesInMins,
-  steps: Steps,
-  tags: Tags,
-  images: Images
-)
-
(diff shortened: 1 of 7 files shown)


common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
```diff
@@ -202,24 +208,31 @@ object CuratedRecipe {
...
   }
 
   def toDBModel(cr: CuratedRecipe): CuratedRecipeDB = {
+
     transform[CuratedRecipe, CuratedRecipeDB](
       cr,
-      "tags" -> getTagNames(cr.tags)
+      "tags" -> getTagNames(cr.tags),
+      "times" -> TimesInMins(
+        Some(cr.times.preparationHours.getOrElse(0.toDouble) * 60.toDouble + cr.times.preparationMinutes.getOrElse(0.toDouble)),
```
LATaylor-guardian: You can move this onto the companion object of `TimesMinAdapted`

LATaylor-guardian:
Other than my comment this looks reasonable 👍

shtukas:
To document the current behaviour:

Original Input:

screen shot 2016-11-14 at 14 00 29

Redisplay:

screen shot 2016-11-14 at 14 01 40

PR #90, opened 2016-11-14
Introduce the TimesInMinsAdapted object

shtukas:


common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
```diff
@@ -69,6 +69,15 @@ case class TimesInMins(
   cooking: Option[Double]
 )
 
+object TimesInMinsAdapted {
+  def preparationTimeInMinutesFromCuratedRecipe(cr: CuratedRecipe): Double = {
```
LATaylor-guardian: You don't need the `FromCuratedRecipe` part. This is observed from the function parameter.

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
```diff
@@ -69,6 +69,15 @@ case class TimesInMins(
...
+    cr.times.preparationHours.getOrElse(0.toDouble) * 60.toDouble + cr.times.preparationMinutes.getOrElse(0.toDouble)
+  }
+  def cookingTimeInMinutesFromCuratedRecipe(cr: CuratedRecipe): Double = {
```
LATaylor-guardian: As above.

shtukas:
You are perfectly right, I am sometimes overly pedantic :)

shtukas:
Done!

LATaylor-guardian:
👍

shtukas:
Thanks!

PR #97, opened 2016-11-15
Add bold to keyboard shortcut keycaps

alastairjardine:
They're a bit hard to read. These are a couple of changes to help that.


shtukas:
:+1:

PR #99, opened 2016-11-15
Implement the updated creation/parsing and verification workflow

shtukas:
This change introduces two utility functions 

1. `getOriginalRecipeStatus(recipeId: String): Option[Status]`, and 
2. `setOriginalRecipeStatus(recipeId: String, s: Status)`

and use them as the basis for implementing the new verification workflow...
`New` (-> `Pending`) -> `Curated` -> `Verified` -> `Finalised` 
...performed by `moveStatusForward(recipeId: String)`


LATaylor-guardian:
Nice! 👍

PR #100, opened 2016-11-15
Writing the function isShowCreation()

shtukas:


common/src/main/scala/com/gu/recipeasy/db/DB.scala
```diff
@@ -164,6 +164,10 @@ class DB(contextWrapper: ContextWrapper) {
     }
   }
 
+  def countRecipesInGivenStatus(status: Status): Long = {
+    contextWrapper.dbContext.run(quote(query[Recipe]).filter(r => r.status == lift(status.toString()))).size
```
LATaylor-guardian: You can omit the () on toString.

ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -30,8 +30,37 @@ class Application(override val wsClient: WSClient, override val conf: Configurat
   }
 
   private def isShowCreation(): Boolean = {
-    // TODO: write the logic to compute the correct answer here
-    true
+
+    // This function indicates whether the user should be redirected to a curation/parsing step or a verification step
```
LATaylor-guardian: This huge comment is unnecessary. Your code should be (and is to me) self documenting.

ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -30,8 +30,11 @@ class Application(override val wsClient: WSClient, override val conf: Configurat
   }
 
   private def isShowCreation(): Boolean = {
-    // TODO: write the logic to compute the correct answer here
-    true
+    val numberOfNewRecipes: Long = db.countRecipesInGivenStatus(New)
```
LATaylor-guardian: Please note that every time we call isShowCreation (i.e every time a user wants to participate) we are making 3 requests to the DB just to decide what view to show them. This is not a problem at the moment, nor may it ever be with the kind of load we are likely to expect but this is something to just be mindful of. No changes are required.

LATaylor-guardian:
👍  when the build passes.

PR #101, opened 2016-11-16
Correct display of time in hours and minutes.

shtukas:
This change makes the preparation and cooking time being correctly displayed online.
The times which are stored in minutes are now correctly reported as hours+minutes.

Also note a small change in the definition of `TimesInMinsAdapted`, to ensure that
the time are reported as integers instead of doubles on the display.

screen shot 2016-11-16 at 11 18 57

screen shot 2016-11-16 at 11 18 03


LATaylor-guardian:
👍

PR #107, opened 2016-11-17
Correct a display bug: last picture caption

shtukas:
Correct a display bug preventing the caption of that last picture of a set of pictures to be seen.

### DIFF
--- a/ui/public/stylesheets/main.css
+++ b/ui/public/stylesheets/main.css
@@ -76,6 +76,7 @@ img, video {
     top: 70px;
     bottom: 50px;
     border-right: 1px solid #d3d3d3;
+    padding-bottom:100px;
 }
 
 .recipe__original__skip-button {


shtukas:
An example of occurrence of the bug (at the time those lines are written) can be found here: https://recipeasy.capi.gutools.co.uk/recipe/4d2a12824ae90735c817d3b38840339e

tomrf1:
👍 if it works!

shtukas:
@tomrf1 Interestingly, I need to push it to prod to have the confirmation because I don't get the pictures on my local. That being said, it's a trivial CSS change so in the worse case, I'll just revert it.

shtukas:
Documenting the current (pre-change) display:

screen shot 2016-11-17 at 10 45 38

shtukas:
Display (post change) :)

screen shot 2016-11-17 at 11 11 12

PR #115, opened 2016-11-21
Update app.scala.html

alastairjardine:
updating progress bar titles to match stages outlined in the FAQ and in the tool


shtukas:
:+1:

PR #119, opened 2016-11-21
Better top message

shtukas:
1. Correct a possible mismatch between the top message and the additional instruction string at the top of a recipe page

2. If somebody visits the curation url, then the recipe is put in `Pending` state (regardless of its previous state).


LATaylor-guardian:
👍

PR #122, opened 2016-11-21
Introduce composite index for unified progress bar

shtukas:
We are now using a composite index for the unified progress. Also the two underlying indices are `Double` (instead of `Int`).


LATaylor-guardian:
👍

ui/app/com/gu/recipeasy/views/app.scala.html
```diff
@@ -1,4 +1,8 @@
-@(title: String, curationIndex: Int, verificationIndex: Int)
+@(title: String, curationIndex: Double, verificationIndex: Double)
+
+@compositeIndex(curationIndex: Double,verificationIndex: Double) = @{
```
shtukas: Changed. Thanks!

PR #123, opened 2016-11-22
Status distribution page

shtukas:
New page shows the distribution recipes per status

### DIFF
--- a/ui/app/com/gu/recipeasy/controllers/Application.scala
+++ b/ui/app/com/gu/recipeasy/controllers/Application.scala
@@ -102,6 +102,18 @@ class Application(override val wsClient: WSClient, override val conf: Configurat
     Ok(views.html.recentactivity(userEventDBs))
   }
 
+  def statusDistribution = AuthAction { implicit request =>
+    val distribution: Map[RecipeStatus, Long] = Map(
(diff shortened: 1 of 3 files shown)


tomrf1:
👍

ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -102,6 +102,18 @@ class Application(override val wsClient: WSClient, override val conf: Configurat
     Ok(views.html.recentactivity(userEventDBs))
   }
 
+  def statusDistribution = AuthAction { implicit request =>
+    val distribution: Map[RecipeStatus, Long] = Map(
+      New -> db.countRecipesInGivenStatus(New),
```
LATaylor-guardian: Again, it's worth being mindful that you are making 6 calls to the DB here when this (if we cared about performance) could probably be achieved with a single query partitioned server side.

shtukas: Yep. I knew you would say that :) , but I didn't mind it this time because this is typically a call myself will make once every hour, or something.

shtukas: That being said, I will be happy to make a change, mostly to learn more about Quill.

PR #124, opened 2016-11-22
Bumping the bootstrap version from 4.0.0-alpha.4 to 4.0.0-alpha.5

shtukas:
In order to implement the new navigation bar, we need the most recent version of Bootstrap.


alastairjardine:
👍

PR #128, opened 2016-11-22
Update multiIngredient.scala.html

alastairjardine:
Adding clarity to the ingredient's widget

`edit` needs a fix, though.


shtukas:
:+1:

PR #131, opened 2016-11-22
Update the skip recipe button with an URL

shtukas:


shtukas:
The button style has been applied to the new anchor

screen shot 2016-11-22 at 15 20 45

alastairjardine:
👍

PR #132, opened 2016-11-22
Update tutorial.scala.html

alastairjardine:


shtukas:
:+1: (checked with AJ)

PR #148, opened 2016-11-24
Introduce a new recipe status

shtukas:
Working towards the split between correctly and incorrectly parsed recipes, introducing a new recipe status: `Ready`. That is going to stand between `New` and `Pending`. In essence `Ready` recipes will be the `New` recipes that have been correctly parsed.


tomrf1:
Makes sense 👍

PR #149, opened 2016-11-24
Correction in Status handling

shtukas:
Noticed this problem at run time after having converted one recipe from `New` to `Ready`


tomrf1:
👍

PR #156, opened 2016-11-24
Unit option chooser: updating display of default (empty) option

shtukas:
This changes allows the default empty option for unit choice to be displayed as `—` while retaining the default `empty` value.


LATaylor-guardian:
👍

PR #157, opened 2016-11-24
Introduce Recipe Preparation

shtukas:
This small change introduces an end-point with an idempotent operation that
is going to convert about 30 recipes from New to Ready.

The purpose is to prepare the database for another commit which is then going
to select them for curation and present them to users instead of recipes
in New status.

The commit after that will upgrade the preparation to include more
recipes (possibly recipes upgraded from the original partial parsing).

### DIFF
--- a/common/src/main/scala/com/gu/recipeasy/models/RecipePreparation.scala
+++ b/common/src/main/scala/com/gu/recipeasy/models/RecipePreparation.scala
@@ -1,0 +1,11 @@
+package com.gu.recipeasy.models
+
+import java.time.OffsetDateTime
+
+import com.gu.recipeasy.db.DB
+
+object RecipePreparation {
+  def selectRecipes(db: DB): List[Recipe] = {
+    db.getRecipes().filter(recipe => ((recipe.ingredientsLists.lists.size >= 5) && (recipe.status == New)))
+  }
+}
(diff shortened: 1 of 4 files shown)


common/src/main/scala/com/gu/recipeasy/models/RecipePreparation.scala
```diff
@@ -0,0 +1,11 @@
+package com.gu.recipeasy.models
+
+import java.time.OffsetDateTime
+
+import com.gu.recipeasy.db.DB
+
+object RecipePreparation {
+  def selectRecipes(db: DB): List[Recipe] = {
+    db.getRecipes().filter(recipe => ((recipe.ingredientsLists.lists.size >= 5) && (recipe.status == New)))
```
tomrf1: This is a one-off operation, and the code will be removed by the subsequent PR

tomrf1:
It would be nice to expand the comments about the various states [here](https://github.com/guardian/recipeasy/blob/master/common/src/main/scala/com/gu/recipeasy/models/Recipe.scala#L32), to make it clear what the new status does

common/src/main/scala/com/gu/recipeasy/models/RecipePreparation.scala
shtukas: Yes. Sorry I should have made that clear. It's just jump-starting the `New` to `Ready` migration

shtukas:
Yes, true. This in planned in the README update that comes next.

PR #161, opened 2016-11-25
Picking up from Ready recipes instead of New recipes

shtukas:
Now that we have `Ready` recipes in the database. Moving to using that subset.


tomrf1:
👍

PR #166, opened 2016-11-25
Overnight process moving Pending recipes to Ready (and no longer to New)

shtukas:
Since recipes are now moving from Ready to Pending, this update bring the recipes
back to Ready.


tomrf1:
👍

PR #170, opened 2016-11-25
Fix scrolling

tomrf1:
1. The left-hand pane now has a vertical scroll bar
2. The tutorial page has some modified styling to disable the scrolling within the panes, as bootstrap tour cannot handle this.


shtukas:
:+1:

PR #174, opened 2016-11-28
Scroll to new ingredients list after pressing 'l'

tomrf1:
At the moment there's no feedback after pressing the `l` key, which creates a new ingredients list. With this change it will scroll to the new list in the right-hand pane.


shtukas:
This is awesome. Thanks. Tested on local. :+1:

PR #177, opened 2016-11-28
Make size our archive clearer in the intro copy

alastairjardine:
making the size of our archive clear in the intro copy


shtukas:
:+1:

PR #179, opened 2016-11-28
Updates in cuisines names

shtukas:
Before:

screen shot 2016-11-28 at 11 52 34

After:

screen shot 2016-11-28 at 11 52 18


tomrf1:
👍

PR #181, opened 2016-11-28
Most insane hack award for Pascal

shtukas:


shtukas:
We don't need the hack anymore :)
See: https://github.com/guardian/recipeasy/pull/182

PR #184, opened 2016-11-28
Improvement in parser: we no longer exclude <em>

shtukas:
This change allows numerous recipes with a ingredient within a `<em>` tag,
for instance
```
<em>2 cups self-raising flour (250g)</em>
```
to be parsed.


tomrf1:
It doesn't really matter, but presumably this means [these comments](https://github.com/guardian/recipeasy/blob/bc22d629d0f63621066b155dcbdd7513eb5a34a9/etl/src/main/scala/etl/RecipeParsing.scala#L91) are now out of date.

👍

PR #188, opened 2016-11-29
Update in display of LHS pictures

shtukas:
1. Reduce (set) the width to 300px.
2. Permute the picture and its caption (captions now on top).


tomrf1:
👍

shtukas:
Thanks!

PR #196, opened 2016-11-29
Disactivate the i-shortcut

shtukas:
The mention of the i-shortcut was already removed from the user interface, but the functionality was still active. This dis-activates it.


tomrf1:
👍

PR #201, opened 2016-11-30
Update in migrateNigelSlaterRecipes

shtukas:
... to avoid the process to un-necessarily run more than once on already migrated recipes


tomrf1:
👍

PR #203, opened 2016-12-01
Cloud formation update with updated Guardian office IP addresses

shtukas:
```
77.91.248.0/21       # London
162.213.134.128/26   # NYC
199.76.32.0/26       # NYC
220.244.211.244/30   # Sydney
125.253.8.136/29     # Sydney
220.244.148.144/28   # Sydney
125.253.8.144/28     # Sydney
```

As per: https://github.com/guardian/ophan-data-lake/blob/8ac770c0c5746a722c25c24f1a6898e57950fdb1/cloudformation/presto.template.yaml#L281-L289

### DIFF
--- a/cloudformation/app.yml
+++ b/cloudformation/app.yml
@@ -192,22 +192,6 @@ Resources:
         - IpProtocol: tcp
           FromPort: 443
           ToPort: 443
-          CidrIp: 220.244.211.244/30 # Sydney
-        - IpProtocol: tcp
-          FromPort: 80
-          ToPort: 80
-          CidrIp: 220.244.211.244/30 # Sydney
-        - IpProtocol: tcp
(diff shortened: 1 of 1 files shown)


tomrf1:
👍 
and yay for comments in CF!

sihil:
The circuits spreadsheet no longer mentions 220.244.211.244/30 or 125.253.8.136/29 so I suspect that these are no longer used by the Guardian (and potentially could have been allocated to other organisations). Otherwise this looks sane. Normal caveats apply of not relying on them as guest wifi falls within these ranges.

shtukas:
@sihil Um... yes I was looking at the Circuits docs after having worked with the Ophan data. So I will remove those ones. Thanks!

PR #204, opened 2016-12-01
Easy Recipeasy

superfrank:
recipeasy

I've introduced SASS and a lot of the mixins used across the site - fonts, colours, breakpoints etc. I've tried to keep the styles just on the main page as buttons etc/styles break on the tools page and I don't quite have enough time to update everything.

This could have probably be done in a better/cleaner way, but as we don't have much time, this is the best I could do. Perhaps if you have time Pascal, you could look at cleaning it up a little?

cc @shtukas


alastairjardine:
Have there been copy changes etc, or is this purely visual updatery?

superfrank:
I removed some copy (from the step 1/step 2), and I think changed a thing in the FAQ but I guess all the copy may need an update in general (so it's more in line with Kath's email).

alastairjardine:
Okay. Let's get this out so we can put in the YouTube vids asap :)

ui/public/stylesheets/shared/components/sass-mq/.gitignore
```diff
@@ -0,0 +1,4 @@
+*.sassc
+.sass-cache
```
mchv: does not seems to wok as there cache files in this PR!

shtukas: I have removed them in the next push

shtukas:
:+1: Thanks @superfrank !

PR #208, opened 2016-12-02
Recipeasy - implementing Caching on progress bar queries.

LATaylor-guardian:
These are now cached for 1 hour and evicted after. The cache has a maximum size of 2.


shtukas:
:+1: From me. Tested locally.

tomrf1:
👍

PR #212, opened 2016-12-02
Removing invisible space in curation page

shtukas:
This remove an invisible char that was rendered as long space.

Before: 

screen shot 2016-12-02 at 13 53 05

After:

screen shot 2016-12-02 at 13 55 15


alastairjardine:
👍

PR #219, opened 2016-12-05
Correction in isShowCreation

shtukas:
When we made the migration from `New` to `Ready` as the starting point of the curation process.
Updating this function was required but forgotten.


tomrf1:
👍

shtukas:
Because Tom was asking, "Creation" is the word used by some people for the step otherwise known as "Curation" because in essence the curation "creates" the structured data 🙃

PR #221, opened 2016-12-05
Updates in the behaviour of accessing already curated recipes on the …

shtukas:
…curation URL.

This change:

1. Makes the header of the read only page better (this page was not accessible to users until now)
2. Add a user event on the access of the read only page
3. Implements the logic for preventing the curation page to be re-accessed outside Ready or Pending status.

### DIFF
--- a/ui/app/com/gu/recipeasy/views/recipe.scala.html
+++ b/ui/app/com/gu/recipeasy/views/recipe.scala.html
@@ -3,28 +3,38 @@
         recipeId: String,
         body: String,
         articleId: String,
-        shouldShowButtons: Boolean,
+        isEditable: Boolean,
         images: List[ImageDB],
         status: RecipeStatus)(implicit messages: play.api.i18n.Messages, request: RequestHeader)
 @import helper.CSRF
 @implicitFC = @{ b4.vertical.fieldConstructor }
 @link = @{ "https://www.theguardian.com/" + articleId }
 @pageTopMessage(status: RecipeStatus) = @{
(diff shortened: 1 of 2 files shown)


shtukas:
New top page of read only page:

read-only

ui/app/com/gu/recipeasy/views/recipe.scala.html
```diff
@@ -3,28 +3,38 @@
...
 @link = @{ "https://www.theguardian.com/" + articleId }
 @pageTopMessage(status: RecipeStatus) = @{
-    status match {
-        case New        => "Creation, pass 1 of 3."
-        case Pending    => "Creation, pass 1 of 3."
-        case Curated    => "Verification, pass 2 of 3."
-        case Verified   => "Final Check, pass 3 of 3."
-        case Finalised  => "Final Check, pass 3 of 3."
-        case _ => ""
+    if(isEditable){
```
LATaylor-guardian: FYI it's encouraged to have a space between `if` and the first bracket and a space between the last bracket and the brace. The same applies with else below. But it's a small thing, don't feel the need to change it in this PR.

LATaylor-guardian:
Looks fine to me 👍

tomrf1:
👍

ui/app/com/gu/recipeasy/views/recipe.scala.html
shtukas: Ok. Cool. I will remember that :)

PR #225, opened 2016-12-06
Correction in curateRecipe

shtukas:
This corrects a little bug that was introduced yesterday.

In the old version, the instruction
```
curatedRecipedEditor(Some(recipe), editable = true)
```
passes to `curatedRecipedEditor` a recipe that carries the `Ready` status, despite the status
having been updated by a previous instruction:
```
db.setOriginalRecipeStatus(recipe.id, Pending)
```

In the new version the instruction:
```
curatedRecipedEditor(db.getOriginalRecipe(id), editable = true)
```
passes to `curatedRecipedEditor` a recipe freshly read from the database and therefore carrying the most
up to date status.


tomrf1:
👍

PR #226, opened 2016-12-06
Better User Events Operation Types

shtukas:
This change:

1. Introduces three new user events `OperationType`s.
2. Override their `toString` method to be (backward) compatible with the strings that have been
   used so far to record user events.

### DIFF
--- a/common/src/main/scala/com/gu/recipeasy/models/UserEvent.scala
+++ b/common/src/main/scala/com/gu/recipeasy/models/UserEvent.scala
@@ -6,9 +6,24 @@ import java.time.OffsetDateTime
 
 sealed trait OperationType
 
-case object Curation extends OperationType
-case object Verification extends OperationType
-case object Confirmation extends OperationType
+case object Curation extends OperationType {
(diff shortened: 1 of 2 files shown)


tomrf1:
Are you going to use the other 3 event case objects?

shtukas:
Actually I am already using them. For instance: https://github.com/guardian/recipeasy/blob/master/ui/app/com/gu/recipeasy/controllers/Application.scala#L35

tomrf1:
I meant Curation/Verification/Confirmation

shtukas:
The strings are just used to write into the database (user event), but as I may soon start querying against them, I wanted them not to be defined more than once.

tomrf1:
ahh I see what's going on now, never mind
👍

shtukas:
> I meant Curation/Verification/Confirmation

Oh. Sorry. My mistake. I am also already using them: https://github.com/guardian/recipeasy/blob/master/ui/app/com/gu/recipeasy/controllers/Application.scala#L128

shtukas:
Ok cool :)

PR #227, opened 2016-12-06
Adding Chinese to the Cuisine list

shtukas:


alastairjardine:
Also to be added:

- Mediterranean 
- Polish

tomrf1:
👍

shtukas:
@alastairjardine I have re-pushed with both Chinese and Mediterranean, but Polish was already there. Cheers,

PR #229, opened 2016-12-06
Update in the style display of read only indicator

shtukas:
Before:

screen shot 2016-12-06 at 08 24 39

After:

screen shot 2016-12-06 at 08 22 02


alastairjardine:
👍

alastairjardine:
Remove the full caps, then 👍

shtukas:
Updated with lower case:

screen shot 2016-12-06 at 09 36 34

PR #231, opened 2016-12-06
Admin HTML templates updates

shtukas:


tomrf1:
not sure what the point of this is, are more pages going to use `layout.scala.html`?
👍

shtukas:
Answering the question :)
I introduced the admin layout to keep things tidy and also to use a mode stable bootstrap than the cutting edge one we use for the curation page.

alastairjardine:
The irony being here that the stable bootstrap is on the bit that's not user facing 😆

PR #232, opened 2016-12-06
The ultimate progress bar computation

shtukas:


shtukas:
Explanation:

When a recipe is put in the database it lands in `New` status, and after that it will go through 5 transitions
1. `New` -> `Ready`
2. `Ready` -> `Pending`
3. `Pending` -> `Curated`
4. `Curated` -> `Verified`
5. `Verified` -> `Finalised`

This means that the progress bar should in fact monitor all possible transitions that the database will go through. 0% means nothing has happened and 100% means nothing else needs to happen.

The total number of transitions is the number of "alive" recipes times 5, where "alive" means not `Impossible`. `Impossible` recipes are meant to remain in that state for ever (there may be exceptions to this but the spirit of "impossible" remains).

So now that we know the total number of transitions, we need to compute how many transitions have been made so far. The way the code does it is `done = total - remain` , this is why the final ratio is like:
```
(possibleMigrationsCount - remainingMigrationCount).toDouble / possibleMigrationsCount
```

The last problem is then: how do we compute the number of remaining transitions ?
Easy: 
A recipe in `New` status has 5 transitions left to do through
A recipe in `Ready` status has 4 transitions left to do through
A recipe in `Pending` status has 3 transitions left to do through
A recipe in `Curated` status has 2 transitions left to do through
A recipe in `Verified` status has 1 transitions left to do through
A recipe in `Finalised` status is in its final state, so no transition left to go though. 

Therefore the formula to compute remaining transitions is
```
val remainingMigrationCount: Long = newCount * 5 + readyCount * 4 + pendingCount * 3 + curatedCount * 2 + verifiedCount * 1 + finalisedCount * 0
```

The last term doesn't add to the sum, but is present so that the mathematical correctness of the code stands out as being obvious.

shtukas:
ps: This implies that the mere fact of releasing new recipes from `New` to `Ready` (something that a parser improvement does) will increase the percentage, but that's the way it should be.

ps: And this also apply that a midnight, when few recipes that had been abandoned by their users are automatically reset from `Pending` to `Ready`, the percentage then decreases by a tiny bit, but then that is also the way it should be :)

tomrf1:
Thanks @shtukas, I now understand what you've done and why.
Of course, this assumes equal weighting for all state transitions, though some will take longer than others.

It gets a 👍 from me, but it might be good to check it matches other team members' definitions of 'progress'.

shtukas:
No worries. The main objective was to be accurate and to make sure that 100% is only displayed when all recipes (beside the `Impossible` ones) are in `Finalised` status (and therefore that there is *really* nothing else to do).

I am going to merge it because I need to do some stats work on the top of that, but I will ask @LATaylor-guardian and @mchv what they think of the computation.

common/src/main/scala/com/gu/recipeasy/db/DB.scala
```diff
@@ -189,17 +189,43 @@ class DB(contextWrapper: ContextWrapper) {
   }
 
   def progressBarRatio(): Double = {
-    val Key = "verificationCompletionRatio"
 
+    // progressBarRatio ratio is a number between 0 and 1
+
+    def pbr(): java.lang.Double = {
```
LATaylor-guardian: If you name this what it is i.e `progressBarRatio` then your code is self documenting and you don't need the comment above. In most having to write comments is a code smell. They should only be necessary to explain something that can't be evident from code.

shtukas: You are right. I have this habit to comment a lot, but admittedly I got it as a protection against evilness while using untyped languages :)

common/src/main/scala/com/gu/recipeasy/db/DB.scala
```diff
@@ -189,17 +189,43 @@ class DB(contextWrapper: ContextWrapper) {
...
+      }
+    }
+
+    val Key = "verificationCompletionRatio"
     val ratio = ProgressCache.get(Key).getOrElse {
-      val touchedRecipes: Long = countRecipesInGivenStatus(Pending) + countRecipesInGivenStatus(Curated) + countRecipesInGivenStatus(Verified) + countRecipesInGivenStatus(Finalised)
-      val activeRecipes: Long = countRecipesInGivenStatus(Ready) + countRecipesInGivenStatus(Pending) + countRecipesInGivenStatus(Curated) + countRecipesInGivenStatus(Verified) + countRecipesInGivenStatus(Finalised)
-      val result: java.lang.Double = touchedRecipes.toDouble / activeRecipes
+      val result: java.lang.Double = pbr()
       ProgressCache.put(Key, result)
```
mchv: That it has this point that you should cast

ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -19,9 +19,8 @@ import models.CuratedRecipeForm._
 class Application(override val wsClient: WSClient, override val conf: Configuration, db: DB, val messagesApi: MessagesApi) extends Controller with AuthActions with I18nSupport with StrictLogging {
 
   def index = AuthAction { implicit request =>
-    val progressBarPercentage: Double = db.progressBarRatio() * 100 // expected to be between 0 and 100
-    val progressBarPercentageRounded = progressBarPercentage - progressBarPercentage % 0.001
-    Ok(views.html.app("Recipeasy", progressBarPercentageRounded))
+    val progressBarPercentage: Double = ( db.progressBarRatio() * 10000 ).toInt.toDouble/100 // expected to be between 0 and 100
```
mchv: `toInt.toDouble` is probably not what you want.
I think you may have the same result with:
```scala
( db.progressBarRatio() * 10000 ) / 100.00
```

common/src/main/scala/com/gu/recipeasy/db/DB.scala
```diff
@@ -189,17 +189,43 @@ class DB(contextWrapper: ContextWrapper) {
...
+      //      Five migrations
+
+      // We apply this to the entire database (including the New) elements, minus the Impossible ones; what we refer to as "alive" recipes below
+
+      val newCount: Long = countRecipesInGivenStatus(New)
+      val readyCount: Long = countRecipesInGivenStatus(Ready)
+      val pendingCount: Long = countRecipesInGivenStatus(Pending)
+      val curatedCount: Long = countRecipesInGivenStatus(Curated)
+      val verifiedCount: Long = countRecipesInGivenStatus(Verified)
+      val finalisedCount: Long = countRecipesInGivenStatus(Finalised)
```
mchv: remove those types

PR #234, opened 2016-12-06
Adding stats to the front page

shtukas:
screen shot 2016-12-06 at 12 17 51

### DIFF
--- a/ui/app/com/gu/recipeasy/controllers/Application.scala
+++ b/ui/app/com/gu/recipeasy/controllers/Application.scala
@@ -21,7 +21,11 @@ class Application(override val wsClient: WSClient, override val conf: Configurat
   def index = AuthAction { implicit request =>
     val progressBarPercentage: Double = db.progressBarRatio() * 100 // expected to be between 0 and 100
     val progressBarPercentageRounded = progressBarPercentage - progressBarPercentage % 0.001
-    Ok(views.html.app("Recipeasy", progressBarPercentageRounded))
+    val usersCount = db.usersCount()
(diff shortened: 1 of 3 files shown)


alastairjardine:
hold for the moment!

shtukas:
No problem.

shtukas:
Better version in progress. Will submit another PR.

ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -21,7 +21,11 @@ class Application(override val wsClient: WSClient, override val conf: Configurat
   def index = AuthAction { implicit request =>
     val progressBarPercentage: Double = db.progressBarRatio() * 100 // expected to be between 0 and 100
     val progressBarPercentageRounded = progressBarPercentage - progressBarPercentage % 0.001
-    Ok(views.html.app("Recipeasy", progressBarPercentageRounded))
+    val usersCount = db.usersCount()
+    val curatedRecipesCount = db.countRecipesInGivenStatus(Curated) + db.countRecipesInGivenStatus(Verified) + db.countRecipesInGivenStatus(Finalised)
+    val finalisedRecipesCount = db.countRecipesInGivenStatus(Finalised)
```
LATaylor-guardian: This is the same call as you're making above. Therefore you can assign the result to a val and use that in both cases removing the need to make 2 DB calls rather than 1.

shtukas: Yes, sure!

shtukas: I didn't think twice about it because I will actually pass a `stats` objects to the view in a latter PR.

PR #235, opened 2016-12-06
Adding a route to extract all recent activities as CSV (text) file

shtukas:
This commit also change the signature of `userEvents` (we now specify the size of the extract we want). We call it with the previous default value (200).


shtukas:
Typical use: 
```
lucille11:~ pascal$ curl localhost:9000/admin/recent-activity-csv
```

But can also be downloaded through a web browser as we set the correct `Content-Type`.

tomrf1:
👍

PR #242, opened 2016-12-08
Compute the function yourParticipation

shtukas:
Notes:

1. The userStats objects's values are strings. So we need to convert them back to Integer for the addition. I will update the structure in a coming PR.

2. @mchv Pointed out that there might be a better way to get the first 3 digits of a decimal number. Currently the operation is abstcted in a separate function: `decimalTo3Digit`

### DIFF
--- a/ui/app/com/gu/recipeasy/views/app.scala.html
+++ b/ui/app/com/gu/recipeasy/views/app.scala.html
@@ -1,8 +1,14 @@
 @import com.gu.recipeasy.models._
 @(title: String, userFirstname:String, progressBarPercentage: Double, generalStats: Map[GeneralStatisticsPoint,Long], userStats: Map[PersonalStatisticsPoint, String])
 
+@decimalTo3Digit(number: Double) = @{
+    ((number)*1000).toInt.toDouble/1000
+}
+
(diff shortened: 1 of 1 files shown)


ui/app/com/gu/recipeasy/views/app.scala.html
```diff
@@ -1,8 +1,14 @@
 @import com.gu.recipeasy.models._
 @(title: String, userFirstname:String, progressBarPercentage: Double, generalStats: Map[GeneralStatisticsPoint,Long], userStats: Map[PersonalStatisticsPoint, String])
 
+@decimalTo3Digit(number: Double) = @{
+    ((number)*1000).toInt.toDouble/1000
```
shtukas: Awesome. Let me try!

LATaylor-guardian: If not, there will be lots of Java classes that have this functionality such as `https://docs.oracle.com/javase/7/docs/api/java/text/NumberFormat.html`

LATaylor-guardian:
Aside from my comment if it proves helpful 👍

shtukas:
@LATaylor-guardian The c-style number formatting did work. Thanks!

PR #243, opened 2016-12-08
Admin update: daily activity breakdown

shtukas:
screen shot 2016-12-08 at 13 26 06

### DIFF
--- a/common/src/main/scala/com/gu/recipeasy/db/DB.scala
+++ b/common/src/main/scala/com/gu/recipeasy/db/DB.scala
@@ -271,6 +271,8 @@ class DB(contextWrapper: ContextWrapper) {
   // ---------------------------------------------
   // User Events
 
+  val isoDateFormat = new java.text.SimpleDateFormat("yyyy-MM-dd")
+
   def countUserEvents(): Long = {
     contextWrapper.dbContext.run(quote(query[UserEvent].schema(_.entity("user_events")))).size
   }
@@ -317,6 +319,27 @@ class DB(contextWrapper: ContextWrapper) {
         query[UserEventDB].schema(_.entity("user_events")).map(event => event.user_email).distinct.size
       )
     )
+  }
+
+  def userEventsDates(): List[String] = {
+    contextWrapper.dbContext.run(
+      quote(
+        query[UserEventDB].schema(_.entity("user_events")).map(event => event.event_datetime)
+      )
+    ).map(datetime => isoDateFormat.format(datetime)).distinct
+  }
+
+  def eventsForDateAndOperationType(date: String, opType: OperationType): List[UserEventDB] = {
+
+    contextWrapper.dbContext.run(
+      quote(
+        query[UserEventDB].schema(_.entity("user_events"))
+      )
+    ).filter(event => (isoDateFormat.format(event.event_datetime) == date) && (event.operation_type == opType.toString()))
+  }
+
+  def dailyActivityDistribution(): List[(String, Int, Int, Int)] = {
+    userEventsDates().sorted.map(date => (date, eventsForDateAndOperationType(date, Curation).size, eventsForDateAndOperationType(date, Verification).size, eventsForDateAndOperationType(date, Confirmation).size))
   }
 
   // ---------------------------------------------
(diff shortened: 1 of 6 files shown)


shtukas:
Note: currently not caching results (for instance figures of past days could be cached), but this should not be a problem because it's not a user facing page...

shtukas:
Updated admin front end

screen shot 2016-12-08 at 13 31 41

common/src/main/scala/com/gu/recipeasy/db/DB.scala
```diff
@@ -319,6 +319,32 @@ class DB(contextWrapper: ContextWrapper) {
     )
   }
 
+  def userEventsDates(): List[String] = {
+    contextWrapper.dbContext.run(
+      quote(
+        query[UserEventDB].schema(_.entity("user_events")).map(event => event.event_datetime)
+      )
+    ).map(datetime => {
```
LATaylor-guardian: Here `.map(datetime => {` the bracket is not necessary and can be written as `.map { datetime =>`

common/src/main/scala/com/gu/recipeasy/db/DB.scala
```diff
@@ -319,6 +319,32 @@ class DB(contextWrapper: ContextWrapper) {
...
+        val format = new java.text.SimpleDateFormat("yyyy-MM-dd")
+        format.format(datetime)
+      }).distinct
+  }
+
+  def eventsForDateAndOperationType(date: String, opType: OperationType): List[UserEventDB] = {
+    val format = new java.text.SimpleDateFormat("yyyy-MM-dd")
```
LATaylor-guardian: You can import `java.text.SimpleDateFormat` seeing as it's being used twice rather than typing the full package name out each time.

common/src/main/scala/com/gu/recipeasy/db/DB.scala
```diff
@@ -319,6 +319,32 @@ class DB(contextWrapper: ContextWrapper) {
...
+    contextWrapper.dbContext.run(
+      quote(
+        query[UserEventDB].schema(_.entity("user_events"))
+      )
+    ).filter(event => (format.format(event.event_datetime) == date) && (event.operation_type == opType.toString()))
+  }
+
+  def dailyActivityDistribution(): List[(String, Int, Int, Int)] = {
```
LATaylor-guardian: This is quite confusing a type `List[(String, Int, Int, Int)]` and would probably be better modelled as a case class.

ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -150,6 +150,11 @@ class Application(override val wsClient: WSClient, override val conf: Configurat
     Ok(views.html.admin.users(usersListing))
   }
 
+  def dailyBreakdown = AuthAction { implicit request =>
+    val dailyActivityDistribution = db.dailyActivityDistribution()
```
LATaylor-guardian: Assigning to a val doesn't really add much here in this case.

ui/app/com/gu/recipeasy/views/admin/dailybreakdown.scala.html
```diff
@@ -0,0 +1,31 @@
...
+                <th>Verification</th>
+                <th>Finalisation</th>
+                <th>Total</th>
+            </tr>
+        </thead>
+        <tbody>
+
+            @for( dayFigures <- dailyActivityDistribution ){
+                <tr>
+                    <th>@{dayFigures._1}</th>
```
LATaylor-guardian: Changing to a case class as mentioned above will also make this code more readable by accessing named members rather than using the tuple accessors.

LATaylor-guardian:
Other than readability and stylistic points 👍

alastairjardine:
shake it so








👍

common/src/main/scala/com/gu/recipeasy/db/DB.scala
shtukas: I have moved the format as val of the class. Updated in next push.

common/src/main/scala/com/gu/recipeasy/db/DB.scala
shtukas: I will make a separate PR for that.

PR #244, opened 2016-12-08
Adding category 'dinner'

shtukas:
This adds the category `dinner` which was missing. Note that we already have the category `Dinner party`. Categories `lunch` and `breakfast` were already present.


LATaylor-guardian:
👍

PR #245, opened 2016-12-08
Cuisines correctly reported after curation

shtukas:
This corrects a bug by which selected cuisines during curation were not shown on the read only recipe view or the verifications steps. They now show up correctly:

screen shot 2016-12-08 at 21 13 00


LATaylor-guardian:
👍

PR #249, opened 2016-12-09
Ingredients quantities atomic step change

shtukas:
This change now allow more natural fractions to be entered (in numeric form). For instance `0.20` and `0.25` can coexist as quantities in the same ingredient list.

screen shot 2016-12-09 at 07 26 16


LATaylor-guardian:
👍

PR #251, opened 2016-12-12
Add the ability to export recipes

mchv:
The associated changes add the ability to export to content API a curated recipe through a new endpoint. The process is the following:

- Retrieve a curated recipe from a `recipe id`
- Retrieve a flexible id using `teleporter` api
- Create an atom model recipe from curated recipe
- Publish atom and update to the original article.

The new endpoint only return a simple text response has I have not yet defined a workflow.
Things to decide are:
- Should be `exported` a new status in the `DB` so we could easily monitor how much recipes we exported?
- Should we allow user to `export` or simply an admin task and export in batch?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -68,7 +68,8 @@ lazy val common = (project in file("common"))
   .settings(flywaySettings)
   .settings(Seq(
       libraryDependencies ++= Seq(
-          "org.jsoup" % "jsoup" % "1.9.2"
+          "org.jsoup" % "jsoup" % "1.9.2",
+          "com.gu" %% "content-api-client" % "10.17"
       )
   ))
 
@@ -77,7 +78,6 @@ lazy val etl = (project in file("etl"))
   .dependsOn(common)
   .settings(Seq(
       libraryDependencies ++= Seq(
-        "com.gu" %% "content-api-client" % "10.5",
         "org.jsoup" % "jsoup" % "1.9.2",
         "org.typelevel" %% "cats-core" % "0.6.1"
       ),
--- a/common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
+++ b/common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
@@ -226,6 +226,9 @@ case class CuratedRecipe(
   images: Images
 )
 
+import com.gu.contentatom.thrift.atom.{ recipe => atom }
+import com.gu.contentatom.thrift._
+
 object CuratedRecipe {
 
   def fromRecipe(r: Recipe): CuratedRecipe = {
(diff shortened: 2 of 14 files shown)


build.sbt
```diff
@@ -68,7 +68,8 @@ lazy val common = (project in file("common"))
   .settings(flywaySettings)
   .settings(Seq(
       libraryDependencies ++= Seq(
-          "org.jsoup" % "jsoup" % "1.9.2"
+          "org.jsoup" % "jsoup" % "1.9.2",
+          "com.gu" %% "content-api-client" % "10.16"
```
LATaylor-guardian: 10.17 now.

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
```diff
@@ -275,5 +278,78 @@ object CuratedRecipe {
...
+            User(email = "off-platform@guardian.co.uk")
+          )
+        )
+      ),
+      revision = 1L
+    )
+
+    val recipeAtom = atom.RecipeAtom(
+      title = cr.title,
+      tags = atom.Tags(
```
shtukas: 1. We are no longer collecting `dietary`, it should not be a part of the model. In fact I am about to remove `dietary` and `holidays` from the main code/model. 
2. What is `celebration` ?
3. Did you mean to use `cuisines` 4 times ?

shtukas: Pascal [ off-plaform ] 8th December at 21:27
> Actually this bug was also present on the old "holidays" selection that we no longer show. (I haven't corrected the bug for "holidays" and what I will actually do is remove the code that used to handle both "holidays" and "dietary" information that we decided not to support; we stopped presenting those two sections on the UI but I had not removed the backend code that handle them... ).

Nathan [ off-plaform ] 9th December at 06.59am
> A point about dietary requirements. We will need to obtain this information after the data capture by inferring it from the ingredients, but I guess removing from recipeasy code is fine!

shtukas: So as discussed, we are not currently collecting `dietary` and `holidays` on the UI, but they are still part of the Elastic Search model. `celebration` means `holidays`. 

`cuisines` was a mistake that @mchv will correct

shtukas: The issue of exposing `dietary` and `holidays` back to the user is in progress....

shtukas:
:+1: Modulo fixing the "cuisines" typo

LATaylor-guardian:
It looks sensible from my side but with lots of integrations the proof will be in the pudding! 👍

PR #253, opened 2016-12-12
Remove bug preventing holidays information to be shown at verification step

shtukas:
This corrects a bug related to the reporting of holidays at verification step. 

A bug which should have been fixed together with this one: https://github.com/guardian/recipeasy/commit/5c1033a0f34334e92bf1b6d1398edc7bbfab188e but wasn't corrected because holidays handling was incorrectly scheduled to be decommissioned.


shtukas:
Note that the front end still doesn't show the drop downs for "Holidays" and "Dietary"

LATaylor-guardian:
👍

PR #256, opened 2016-12-15
Update capi model for recipe

mchv:
Images are not yet exported and we need to implement quantity ranges but this the start


shtukas:
:+1:

PR #257, opened 2016-12-16
Allow to publish to Content API on GET

mchv:
I will put it back to `POST` and create a form later,  but I need to bypass the CSRF token filter for my first tests.


LATaylor-guardian:
There's no javascript calling this route that could cause problems is there?

Apart from that 👍

mchv:
No at all it is just for testing for the moment so hide it in the UI

PR #258, opened 2016-12-16
Add the ability to assume role and publish to kinesis streams in flexible conte

mchv:
This is necessary for publishing a `recipe` - See #251


mchv:
I have applied to `PROD`

mchv:
I am still get the error `Unable to load AWS credentials from any provider in the chain` and I am not sure now what I am missing.

PR #260, opened 2016-12-19
Recipeasy: Fix publishing permissions.

LATaylor-guardian:
- When using an STS Role, a STSAssumeRoleSessionCredentialsProvider needs to be available in the AWSCredentialsProviderChain. @mchv FYI.


tomrf1:
👍

PR #275, opened 2016-12-20
Adding Soup as new category

shtukas:


shtukas:
Verbal +1 from @tomrf1, who had to leave, but reviewed it :)

PR #279, opened 2016-12-22
Introduce ranges for ingredient lists

shtukas:
This change enable range description for each ingredient. For this

1. We change `case class DetailedIngredient`
2. Modify the User Interface with two additional fields for ingredients
3. Echo the range in the receipe atom where a Double to Short is performed.

### DIFF
--- a/common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
+++ b/common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
@@ -71,6 +71,15 @@ object CuratedRecipe {
     })
   }
 
+  def atomRangeFromOptionalBounds(maybeFrom: Option[Double], maybeTo: Option[Double]): Option[atom.Range] = {
+    for {
+      from <- maybeFrom
+      to <- maybeTo
+    } yield {
+      atom.Range((from * 100).toShort, (to * 100).toShort)
+    }
+  }
+
   def toAtom(r: Recipe, cr: CuratedRecipe): Atom = {
 
     val contentChangeDetails = ContentChangeDetails(
@@ -129,6 +138,7 @@ object CuratedRecipe {
               item = ingredient.item,
               comment = ingredient.comment,
               quantity = ingredient.quantity,
+              quantityRange = atomRangeFromOptionalBounds(ingredient.quantityRangeFrom, ingredient.quantityRangeTo),
               unit = ingredient.unit.map(_.abbreviation)
             ))
         )),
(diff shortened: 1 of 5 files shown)


shtukas:
screen shot 2016-12-22 at 14 33 24

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipeTypes.scala
```diff
@@ -51,7 +51,15 @@ object DetailedIngredientsLists {
       ingredients.lists.map(r => new DetailedIngredientsList(r.title, rawToDetailedIngredients(r.ingredients))))
   }
   private def rawToDetailedIngredients(ingredients: Seq[String]): Seq[DetailedIngredient] = {
-    ingredients.map(i => new DetailedIngredient(None, None, "", None, i))
+    ingredients.map(i => new DetailedIngredient(
```
LATaylor-guardian: `new` is not needed for case classes.

shtukas: Thanks you are right!

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
```diff
@@ -71,6 +71,16 @@ object CuratedRecipe {
...
   }
 
+  def atomRangeFromOptionalBounds(maybeFrom: Option[Double], maybeTo: Option[Double]): Option[atom.Range] = {
+    for {
+      from <- maybeFrom
+      to <- maybeTo
+    } yield {
+      atom.Range((from * 100).toShort, (to * 100).toShort)
+    }
+    None
```
LATaylor-guardian: Why does this return None? The for comprehension will return an Option.

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipeTypes.scala
LATaylor-guardian: When a map function stretches over multiple lines its `ingredients.map  { ` rather than `ingredients.map( `

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
shtukas: Sorry. Yes.

`atom.Range((from * 100).toShort, (to * 100).toShort)` was initially commented out while I was sorting the atom problem, and initially returned `None` to compile and when I corrected the atom I forgot to remove the `None`

LATaylor-guardian:
Other than my comments 👍

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipeTypes.scala
shtukas: Corrected. Update coming in.

PR #280, opened 2017-01-03
Improvements in the display of the range

shtukas:
In this commit we modify/improve the display of the ingredient list range. A checkbox
now controls the toggling of single measurement versus range measurement.

**Before**

screen shot 2017-01-03 at 15 01 54

**After**

screen shot 2017-01-03 at 15 02 34

screen shot 2017-01-03 at 15 02 49

Also, removal of the "_" standing for "no unit" in the serve information.

### DIFF
--- a/ui/public/stylesheets/main.css
+++ b/ui/public/stylesheets/main.css
@@ -725,6 +725,14 @@ img, video {
   margin-bottom: 1px;
   margin-right: 10px; }
 
+.ingredient__single-measurement-field {
+
+}
+
+.ingredient__range-measurement-field {
+  display:none
+}
+
 .form-hidden {
   background: transparent;
   border: none; }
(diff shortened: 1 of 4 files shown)


ui/public/stylesheets/main.css
```diff
@@ -725,6 +725,14 @@ img, video {
   margin-bottom: 1px;
   margin-right: 10px; }
 
+.ingredient__single-measurement-field {
```
tomrf1: does this have any effect?

shtukas: Um..... yes. Well, it is used anyway, in "editForm.js"

```
$(this).parent().parent().find(".ingredient__single-measurement-field").hide()
```

tomrf1: the class is used, but the empty css block seems odd...

tomrf1:
👍 This looks much clearer to me now!

shtukas:
Cool! Just waiting for @alastairjardine 's seal of approval and will deploy :)

alastairjardine:
Could the input boxes be smaller? They seem quite large. It would be nice if they could be tighter. 

Same probably goes for the single input box...

alastairjardine:
also, does it look better without space between the fields and the en-dash

shtukas:
So... the left hand space came from the Boostrap CSS surrounding the number fields, so i added a space on the right hand side to compensate. But I can try and reduce both sides with some magic. BRB...

shtukas:
Just saw the request for the boxes size. So the standard size for the single measurement was 100px and I reduced it to 80px for each measurements of a range. I will adjust the sizes further in the next PR.

alastairjardine:
👍

PR #281, opened 2017-01-03
Stylistic updates in Ingredient item display

shtukas:
- width of a single measurement is now 80px
- placeholder for single measurement is now "amount"
- width for range component is now 60px
- the separator for range is now more tightly displayed.

screen shot 2017-01-03 at 16 21 37


tomrf1:
👍

PR #284, opened 2017-01-04
In this commit

shtukas:
1. We allow the users to enter Servings quantities with two new units: litres and kilograms

2. We update the `toAtom(r: Recipe, cr: CuratedRecipe): Atom` function with an automatic conversion of
   litres and kilograms ( quantities and units ) to millilitres and grams

3. Note that the return value of atomServeQuantity is always Short
   therefore, in the case of millilitres and grams, we convert the Double to Short

The strings "litres", "kilograms", "millilitres" and "grams" will be made into a type in the next PR :)


tomrf1:
👍 Looks fine, but you should definitely model units properly soon

shtukas:
Cool ! :)

PR #285, opened 2017-01-04
Refactoring

shtukas:
In this commit I un-hardcode the strings used for serving units
( as I promised I would do here: https://github.com/guardian/recipeasy/pull/284 )


common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
```diff
@@ -109,18 +104,13 @@ object CuratedRecipe {
...
-        The code will be made more robust in a later PR
-    */
     portion match {
       case QuantityType => {
         maybeUnit match {
           case Some(unit) => {
             unit match {
-              case "kilograms" => Some("grams")
-              case "litres" => Some("millilitres")
+              case PortionUnitKilograms.name => Some("grams")
```
LATaylor-guardian: If the name matches you can use for the Some value aswell, no?

LATaylor-guardian:
👍

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
shtukas: Yep. Done. Pushed. Cheers!

PR #286, opened 2017-01-04
New logic behind the `/recipe/curate-or-verify` end point

shtukas:
In this change we replace the choice originally made by function `isShowCreation()`
with a new logic. A user is always first given a verification (or confirmation) step
assuming that this doesn't lead to the user verifying or confirming a recipe they
had already curated or verified.

For this we introduce the database function `getUserSpecificRecipeForVerificationStep(userEmail: String): Option[Recipe]`
which uses information taken from the user_events table.


tomrf1:
👍

PR #289, opened 2017-01-05
Refactoring (removing hard coded strings, notably in Database Queries)

shtukas:
In this change (the first of two refactoring changes), we apply the method used in
this commit: https://github.com/guardian/recipeasy/commit/d5ce6dc1ceec19855b9d56993144ea1605d5af84
to make the code more robust and less amendable to accidental breaks.

Note that in the case of `sealed trait OperationType` we had already used the type itself
but relied on function `toString` instead of a value and this wasn't working with Quill.
This change therefore moves `OperationType`'s `toString` functions to names.


tomrf1:
👍

PR #290, opened 2017-01-05
Refactoring: Types renaming

shtukas:
Because of linguistic similarities between the names of RecipeStatus and OperationType, the code is
sometimes difficult to follow. For instance `case object Curation` and `case object Curated` sometimes
cause confusion.

This change, ( the second part of https://github.com/guardian/recipeasy/commit/9d6a10bb11ba25da851285e0b6c374d1094bb98f )
solves that problem.

As an example, the code of the following function makes more sense, and hilights better that both user events and recipes status are involved in the resuling query
```
  def getUserSpecificRecipeForVerificationStep(userEmail: String): Option[Recipe] = {
    val recipeIdsAlreadyTouchedByThisUser = quote {
      query[UserEventDB].schema(_.entity("user_events"))
        .filter(event => event.user_email == lift(userEmail))
        .filter(event => ((event.operation_type == lift(UserEventCuration.name)) || (event.operation_type == lift(UserEventVerification.name)) || (event.operation_type == lift(UserEventConfirmation.name))))
        .map(event => event.recipe_id)
    }
    val q2 = quote {
      query[Recipe]
        .filter(r => ((r.status == lift(RecipeStatusCurated.name)) || (r.status == lift(RecipeStatusVerified.name))))
        .filter(r => !recipeIdsAlreadyTouchedByThisUser.contains(r.id))
        .take(1)
    }
    contextWrapper.dbContext.run(q2).headOption
  }
```

### DIFF
--- a/common/src/main/scala/com/gu/recipeasy/db/DB.scala
+++ b/common/src/main/scala/com/gu/recipeasy/db/DB.scala
@@ -22,14 +22,14 @@ class DB(contextWrapper: ContextWrapper) {
   private implicit val decodePublicationDate = mappedEncoding[Date, OffsetDateTime](d => OffsetDateTime.ofInstant(d.toInstant, ZoneOffset.UTC))
   private implicit val encodeStatus = mappedEncoding[RecipeStatus, String](_.toString())
   private implicit val decodeStatus = mappedEncoding[String, RecipeStatus](d => d match {
-    case "New" => New
-    case "Ready" => Ready
-    case "Pending" => Pending
-    case "Curated" => Curated
-    case "Verified" => Verified
-    case "Finalised" => Finalised
-    case "Impossible" => Impossible
-    case _ => Impossible
+    case "New" => RecipeStatusNew
+    case "Ready" => RecipeStatusReady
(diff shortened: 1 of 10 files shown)


common/src/main/scala/com/gu/recipeasy/db/DB.scala
```diff
@@ -22,14 +22,14 @@ class DB(contextWrapper: ContextWrapper) {
...
   private implicit val decodeStatus = mappedEncoding[String, RecipeStatus](d => d match {
-    case "New" => New
-    case "Ready" => Ready
-    case "Pending" => Pending
-    case "Curated" => Curated
-    case "Verified" => Verified
-    case "Finalised" => Finalised
-    case "Impossible" => Impossible
-    case _ => Impossible
+    case "New" => RecipeStatusNew
```
LATaylor-guardian: Rather than prefixing each one with `RecipeStatus`, you could create an object to hold them in called `RecipeStatus`. That way you would access them like `RecipeStatus.new`.

tomrf1:
👍

PR #291, opened 2017-01-05
Echoing curation user identity on the curation page

shtukas:
screen shot 2017-01-05 at 16 26 55

### DIFF
--- a/ui/app/com/gu/recipeasy/controllers/Application.scala
+++ b/ui/app/com/gu/recipeasy/controllers/Application.scala
@@ -34,7 +34,7 @@ class Application(override val wsClient: WSClient, override val conf: Configurat
   def viewRecipe(id: String) = AuthAction { implicit request =>
     val recipe = db.getOriginalRecipe(id)
     db.insertUserEvent(UserEvent(request.user.email, request.user.firstName, request.user.lastName, id, UserEventAccessRecipeReadOnlyPage.name))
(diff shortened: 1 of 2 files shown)


ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -34,7 +34,7 @@ class Application(override val wsClient: WSClient, override val conf: Configurat
   def viewRecipe(id: String) = AuthAction { implicit request =>
     val recipe = db.getOriginalRecipe(id)
     db.insertUserEvent(UserEvent(request.user.email, request.user.firstName, request.user.lastName, id, UserEventAccessRecipeReadOnlyPage.name))
-    curatedRecipedEditor(recipe, editable = false)
+    curatedRecipedEditor(recipe, editable = false, userEmail = request.user.email, userFirstName = request.user.firstName, userLastName = request.user.lastName)
```
LATaylor-guardian: Given all the properties are from the user object, would it be easier to only pass through the user object?

LATaylor-guardian:
Once tests pass 👍

ui/app/com/gu/recipeasy/controllers/Application.scala
shtukas: Um.... I thought of that, but I decided to reproduce what I did for the main page. I know it's not as beautiful, but the type of the user itself is... funny.

PR #292, opened 2017-01-05
Correction of commit ae3e48e3fd6ae7dce1c317db9b4e07c1124142c5

shtukas:
The previous commit is incorrect because the displayed name is not meant to be the current user but the
user who has performed the curation.
(This mistake wasn't apparent at the moment the PR was done because on my computer both users are teh same).

This commit introduces the database function `getFirstCurationEventForRecipe(recipeId: String): Option[UserEventDB]`
which extracts an optional UserEventDB.

It is used by `getCurationUser(recipeId: String, db: DB): CurationUser`
which given a recipeId and a database object returns a CurationUser. Note that a
CurationUser is always returned, even if the previous function returned null, this to keep the code of the
Application controller and the code of the view more simple.


shtukas:
Auto :+1: as there is nobody else left in the office

tomrf1:
👍 though I think you meant "returned None" rather than "returned null"!

shtukas:
Yes sorry :)

shtukas:
Thanks!

PR #293, opened 2017-01-06
Correction in post curation form redirection logic

shtukas:
This commit corrects the redirection logic after the submission of a curation form.
The user should be redirected to the curateOrVerify end point.

Also, code cleaning: `finalCheckRecipe` and `curateOneRecipeInNewStatus` are not used.


tomrf1:
👍

PR #295, opened 2017-01-06
Randomisation of the selection of a recipe to verify

shtukas:


shtukas:
Note: When you curate a recipe (here defined as just accessing the page), it moves from `Ready` to `Pending` and only when you submit the form it goes from `Pending` to `Curated`. The `Pending` step prevents two users to be given the same recipe to curate (as they would override each other's changes). (And as we all know, recipes in `Pending` status are sent back to `Ready` during the night)

Unfortunately there isn't (yet) a similar process for the migration from `Curated` to `Verified` or `Verified` to `Confirmed`. Moreover the selection of the recipe to verify (or confirm) is deterministic, the consequence of the two above is that two users can be given the same recipe to verify (or confirm).

This change fix the problem by randomising the choice of the recipe

shtukas:
Note: The randomisation is done in memory and not at database level, because I don't think Quill supports that

tomrf1:
👍 as a temporary solution

PR #299, opened 2017-01-09
Renaming RecipeStatusPending into RecipeStatusPendingPostReady

shtukas:


tomrf1:
`RecipeStatusPendingCuration`?

shtukas:
Yep, I thought of that, but didn't know if it sounded better. Give me a moment, I will change it :)

tomrf1:
It makes more sense to me, others may disagree! 'Pending' implies an action is about to take place, so it seems reasonable to name the pending action. Presumably there'll be a `RecipeStatusPendingVerification` too.
Naming things is hard!

shtukas:
No worries, your naming is much better than mine! I have pushed the update :)

tomrf1:
👍

PR #302, opened 2017-01-09
Update in HouseKeepingJob

shtukas:
Now that we have more than one "pending" status, the overnight database function that sends pending recipes to their previous state needs updating.

### DIFF
--- a/common/src/main/scala/com/gu/recipeasy/db/DB.scala
+++ b/common/src/main/scala/com/gu/recipeasy/db/DB.scala
@@ -158,9 +158,13 @@ class DB(contextWrapper: ContextWrapper) {
     contextWrapper.dbContext.run(quote(query[Recipe]).filter(r => (r.status == lift(RecipeStatusCurated.name) || r.status == lift(RecipeStatusVerified.name))).sortBy(r => r.publicationDate)(Ord.desc).take(1)).headOption
   }
 
-  def resetOriginalRecipesStatus(): Unit = {
-    val a = quote(query[Recipe].filter(_.status == "Pending").update(_.status -> "Ready"))
+  def resetOriginalRecipesInPendingStatuses(): Unit = {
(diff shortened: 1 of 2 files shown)


LATaylor-guardian:
👍

common/src/main/scala/com/gu/recipeasy/db/DB.scala
```diff
@@ -158,9 +158,13 @@ class DB(contextWrapper: ContextWrapper) {
     contextWrapper.dbContext.run(quote(query[Recipe]).filter(r => (r.status == lift(RecipeStatusCurated.name) || r.status == lift(RecipeStatusVerified.name))).sortBy(r => r.publicationDate)(Ord.desc).take(1)).headOption
   }
 
-  def resetOriginalRecipesStatus(): Unit = {
-    val a = quote(query[Recipe].filter(_.status == "Pending").update(_.status -> "Ready"))
+  def resetOriginalRecipesInPendingStatuses(): Unit = {
+    val a = quote(query[Recipe].filter(_.status == lift(RecipeStatusPendingCuration.name)).update(_.status -> lift(RecipeStatusReady.name)))
```
tomrf1: Rather than assigning to these weirdly-named values, can we just do something like:
```
contextWrapper.dbContext.run {
  quote(query[Recipe].filter(_.status == lift(RecipeStatusPendingCuration.name)).update(_.status -> lift(RecipeStatusReady.name)))
}
```

Or even stick them in a list and do `queries.map(contextWrapper.dbContext.run)`

tomrf1: never mind, too late :)

LATaylor-guardian: @tomrf1 is quite right - the assignment is unnecessary.

shtukas: Ok, ok. I will correct that, in few minutes.... 
If fact the style of db calls in that file is not consistent, I will correct that as well :)

PR #303, opened 2017-01-09
Update in progress bar computation (to include new statuses)

shtukas:


shtukas:
Note: This will not change the current value, just aligned the computation with the increase of pending statuses.

tomrf1:
👍

PR #304, opened 2017-01-09
Refactoring in the DB.scala file

shtukas:
If the quoted action is one single expression then we do not store in in a
`val` before passing it to the `run` method, otherwise we call the last expression "action"

This was prompted by the comments of https://github.com/guardian/recipeasy/pull/302


tomrf1:
👍 much nicer

shtukas:
:)

PR #305, opened 2017-01-09
Implementation of the new status workflow.

shtukas:
In this commit we add the logic to manipulate the two new recipe statuses that we introduced
here: https://github.com/guardian/recipeasy/commit/efc7be54bc39ade37be92b970fb95880cbfe4949

### DIFF
--- a/common/src/main/scala/com/gu/recipeasy/db/DB.scala
+++ b/common/src/main/scala/com/gu/recipeasy/db/DB.scala
@@ -182,11 +182,21 @@ class DB(contextWrapper: ContextWrapper) {
     )
   }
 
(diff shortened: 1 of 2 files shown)


shtukas:
Note: The random choice that was introduced here ( https://github.com/guardian/recipeasy/commit/b4a6b18e8fc5fba9999ffc9d5b77cb630de7ee24 ) was meant to reduce the probability of two users performing a verification step on the same recipe at the same time. The current commit fixes the underlying problem by making it impossible for two users to look at the same recipe for verification ( unless they both access the page, a bit before and a bit after midnight ), this being said, there should not be any reason to actually give up the random choice itself.

tomrf1:
👍

PR #306, opened 2017-01-09
New admin page for user activity

shtukas:


ui/app/com/gu/recipeasy/views/admin/index.scala.html
```diff
@@ -4,6 +4,7 @@
     <h2>Recipeasy Admin Area</h2>
     <ul>
         <li><a href="@routes.Application.recentActivity">Recent activity</li>
+        <li><a href="@routes.Application.recentUserActivity("alice.skywalker@guardian.co.uk")">Recent activity of a specific user</li>
```
mchv: why do you had a specific user `here`?

shtukas: So......... I want the url to appear on the admin landing page (so that we can quickly remember what it is), but without singling out a single user. "Alice Skywalker" is not a real user (not a real employee), so her activity is empty.

I did this because otherwise a link from the curation page would open this new admin page, but I don't want users to even know about the existence of the page.

The intended use of this page is @alastairjardine click on it and manually replace the fake email address with the email address of the user he is interested in. 

Alternatively, I can add a link on the curation page, but I will have to make that link appear only to the admins who actually can see that page.

Either way works for me :)

shtukas:
screen shot 2017-01-09 at 16 11 11

shtukas:
View with a non fake user:

screen shot 2017-01-09 at 16 12 17

ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -18,6 +18,16 @@ import models.CuratedRecipeForm._
 
 class Application(override val wsClient: WSClient, override val conf: Configuration, db: DB, val messagesApi: MessagesApi) extends Controller with AuthActions with I18nSupport with StrictLogging {
 
+  val AdminWithClearanceUsers : List[String] = List(
+    "alastair.jardine@guardian.co.uk",
+    "nathan.good@guardian.co.uk",
+    "pascal.honore@guardian.co.uk"
```
mchv: Please add everyone in the `off-platform` team

shtukas: On it!

mchv: Please not as well that the repo is public so it would be better to check if the person if from a group like in `janus`.

shtukas: Started working on that. Cheers,

ui/app/com/gu/recipeasy/views/admin/index.scala.html
shtukas: Something just occurred to me. The default user should simply be the current user. If the current user in admin then the page will display fine (their own activity), after which they can put whichever email address they want, otherwise the page will 404.

Update coming.

ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -18,6 +18,16 @@ import models.CuratedRecipeForm._
 
 class Application(override val wsClient: WSClient, override val conf: Configuration, db: DB, val messagesApi: MessagesApi) extends Controller with AuthActions with I18nSupport with StrictLogging {
 
+  val AdminWithClearanceUsers: List[String] = List(
+    "alastair.jardine@guardian.co.uk",
+    "nathan.good@guardian.co.uk",
+    "pascal.honore@guardian.co.uk"
```
adamnfish: Definitely be good to take these email addresses from settings that aren't in this repository.

There are several options. This can be part of the application config, you could use the Google Group Checker provided in Play-GoogleAuth, or you could use some other service to provide the list. My understanding is that other admin tools we provide use Google groups to achieve this, we can chat IRL if you like - I can point you at the right people.

ui/app/com/gu/recipeasy/views/admin/index.scala.html
```diff
@@ -1,9 +1,10 @@
-@()(implicit messages: play.api.i18n.Messages, request: RequestHeader)
+@(loggedInUserEmailAddress: String)(implicit messages: play.api.i18n.Messages, request: RequestHeader)
 @layout("Recipeasy Admin") {
 
     <h2>Recipeasy Admin Area</h2>
     <ul>
         <li><a href="@routes.Application.recentActivity">Recent activity</li>
+        <li><a href="@routes.Application.recentUserActivity(loggedInUserEmailAddress)">Recent activity of a specific user</li>
```
adamnfish: Would this not be better provided as a form? Allowing the user to provide the email address would remove the choice of "what's the email address we should show first", as well as simplifying the usage.

ui/conf/routes
```diff
@@ -23,6 +23,7 @@ GET        /recipe/publish/:recipeId                                      contro
 GET        /admin                                                         controllers.Application.adminLandingPage
 GET        /admin/recent-activity                                         controllers.Application.recentActivity
 GET        /admin/recent-activity-csv                                     controllers.Application.recentActivityCSV
+GET        /admin/recent-user-activity/:userEmail                         controllers.Application.recentUserActivity(userEmail)
```
adamnfish: Related to my point above about choosing the userEmail, you could choose to provide GET and POST handlers on this endpoint. The former might show a form for selecting the user and the latter display the page showing that user's activity (this one being what you have now).

ui/app/com/gu/recipeasy/controllers/Application.scala
shtukas: Yep, we are currently setting up a google group to grant the access. Update tomorrow....

ui/app/com/gu/recipeasy/views/admin/index.scala.html
shtukas: Well... if you put it like this :) I will do that tomorrow :)

ui/conf/routes
shtukas: :+1:

PR #307, opened 2017-01-10
Update in DB.scala's private implicit val decodeStatus

shtukas:
Filling up missing case clauses not previously caught.


tomrf1:
👍

PR #309, opened 2017-01-10
Resolving a bug in the creation of new ingredients when using JavaScript

shtukas:
The unit will now we correctly created with a neutral value.


shtukas:
The way the (original) javascript code creates new ingredient is simply to (1) copy a previous ingredient, (2) clean it from its values, (3) attempts to populate it using the information coming from the selected text.

The cleaning step was working for the `input` part of a ingredient, but not the `select` part. The consequence of this bad cleaning is that the unit that was displayed was simply copied from a previous ingredient.

shtukas:
Note: this bug only manifests itself when the javascript code fails to find a unit for the newly created ingredient

tomrf1:
👍

PR #310, opened 2017-01-10
Allowing 0.125 as ingredient quantity

shtukas:


tomrf1:
might be nice to move it into a `val`? 👍

shtukas:
Um.......Not a bad idea... If I ever need to change it again. I will :)

PR #318, opened 2017-01-12
Change in leaderboard comparison function

shtukas:
This small change prevents two objects that are not actually the same object
to be rendered equal by the comparison function.

The order has to be an antisymmetric relation
https://en.wikipedia.org/wiki/Antisymmetric_relation


shtukas:
:+1: as per discussion with @mchv

tomrf1:
👍

PR #319, opened 2017-01-12
Display improvements in the leaderboard (timings)

shtukas:
Here we do not display timings equal to 0, which actually mean "no data" and
we display the numbers up to two decimals.


shtukas:
screen shot 2017-01-12 at 14 10 44

tomrf1:
👍

PR #320, opened 2017-01-12
Refactorings

shtukas:
Note: `.drop(1)`

```
scala> List(1).drop(1)
res0: List[Int] = List()

scala> List().drop(1)
res1: List[Nothing] = List()
```

Leads to a nicer code


LATaylor-guardian:
👍

PR #321, opened 2017-01-12
Recipeasy - fix extremely damaging bug.

LATaylor-guardian:


shtukas:
:+1: ( because Recipeasy has been so unfair to you ^^ )

LATaylor-guardian:
My good name has been besmirched by this slanderous tool.

shtukas:
One the other hand you could just do one curation and you will go to position at least 128 :)

PR #322, opened 2017-01-13
Cloud formation change for recipeasy-service-account.json

shtukas:
This bring the recipeasy-service-account.json file to the box for future use.


shtukas:
Cloud formation stack already applied and deployment has been tested.

tomrf1:
👍

PR #326, opened 2017-01-16
Introduce user activity page

shtukas:


ui/app/com/gu/recipeasy/controllers/Application.scala
```diff
@@ -13,6 +13,10 @@ import com.gu.recipeasy.models._
 import com.gu.recipeasy.views
 import models._
 import models.CuratedRecipeForm._
+import auth._
+
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
```
LATaylor-guardian: You can use the play execution context: `play.api.libs.concurrent.Execution.Implicits._`. https://www.playframework.com/documentation/2.5.x/ThreadPools

LATaylor-guardian:
Other than my comment 👍

shtukas:
Added item on the global activity page

screen shot 2017-01-16 at 17 35 59

ui/app/com/gu/recipeasy/controllers/Application.scala
shtukas: Actually I don't need those imports. Cheers,

PR #327, opened 2017-01-17
Publish curated recipes

annebyrne:
Change publishing so that curated recipes are immediately pushed to CAPI rather than waiting for 3 step verification process. To allow greater amount of data for testing.

PR #330, opened 2017-01-17
Implements the atom publishing at each form submission

annebyrne:
Change publishing so that curated recipes are immediately pushed to CAPI rather than waiting for 3 step verification process. To allow greater amount of data for testing.


shtukas:
Replaces https://github.com/guardian/recipeasy/pull/327

shtukas:
:+1:

PR #331, opened 2017-01-18
Introduce atom last modified date using current time

annebyrne:


shtukas:
Looks good but I will let @mchv +1 it :)

build.sbt
```diff
@@ -20,7 +20,8 @@ val commonSettings = Seq(
     "com.github.ben-manes.caffeine" % "caffeine" % "2.3.5",
     "com.google.code.findbugs" % "jsr305" % "3.0.1", // required to prevent Caffeine causing compile to fail given -Xfatal-warnings flag.
     "com.amazonaws" % "amazon-kinesis-client" % "1.6.2",
-    "org.scalatest" %% "scalatest" % "2.2.6" % Test
+    "org.scalatest" %% "scalatest" % "2.2.6" % Test,
```
LATaylor-guardian: Small thing, but stylistically test dependencies should always be at the bottom of your dependency list.

common/src/main/scala/com/gu/recipeasy/models/CuratedRecipe.scala
```diff
@@ -129,7 +130,7 @@ object CuratedRecipe {
       ),
       published = Some(
         ChangeRecord(
-          date = r.publicationDate.toInstant.toEpochMilli,
+          date = DateTime.now.toInstant.getMillis,
```
LATaylor-guardian: @mchv  @annebyrne Don't we want the publication date to be that of when the original recipe article was created?

shtukas: No, it was a request from @mchv that we set it to the current time. (At least that's what I understood...)

mchv: I was thinking the the publication date of the atom should be `updated` when we publish a new version of the atom. The other option is to use `last-modified` but I am unsure if the difference is worth as we had on a legacy on article, what do you think?

shtukas: I vote for having the atom behave independently to the article. Mostly because one day we may have a atoms which do not have corresponding articles and in this case the coupling would not make sense. 

But others may disagree :)

mchv: @LATaylor-guardian @clloyd This is the [current documentation](https://github.com/guardian/content-atom/blob/master/thrift/src/main/thrift/contentatom.thrift#L38-L45) 
```thrift
  /** the latest change to the content atom */
  1: optional shared.ChangeRecord lastModified

  /** the atom creation event */
  2: optional shared.ChangeRecord created

  /** the atom publication event (if published) */
  3: optional shared.ChangeRecord published
```
Is `published` the first publication date or should it be updated each time the atom is `published`?

mchv:
@annebyrne @shtukas following discussion with @LATaylor-guardian we agreed on:

- `published` should be the original article publication time (as it is on master)
- `lastModified` should be set as the now time (as you did in this PR)

annebyrne:
@mchv Have introduced changes based on your discussion about publication vs lastModified.

mchv:
👍

PR #332, opened 2017-01-18
Make tag fields lowercase in atom

tomrf1:
This makes things easier for Concierge, which will match on fields in elasticsearch which have not been tokenized (e.g "south america").


shtukas:
:+1:

PR #333, opened 2017-01-18
Introduce a new user event UserEventAccessFrontPage

shtukas:
This introduce a new user event for an event that wasn't recorded
yet: a user accessing the front page (which would have been useful to me
this morning).

The admin pages are updated accordingly (it is the first event without an associated
recipe id).

screen shot 2017-01-18 at 12 14 48


tomrf1:
👍

Issue #266, opened 2016-12-20
A [natural measurement] of


shtukas:
In "A [natural measurement] of", "of" should be stripped from [thing] contents:

![image](https://cloud.githubusercontent.com/assets/6035518/21354619/bdadfe92-c6c2-11e6-8ecb-24208f5abcdc.png)

shtukas:
Note. This issue is solved when the ingredient list is built using the front end shortcuts, but the original ETL parser might still need to be updated.

PR #335, opened 2017-01-18
Clean ingredient item field of unnecessary inherited 'of' from unit measurement

annebyrne:
Resolves https://github.com/guardian/recipeasy/issues/266


shtukas:
:+1:, but I will ask @tsop14 's second opinion as she is much better than me on this things :)

tsop14:
Nice (assuming will always be lowercase) 👍

PR #336, opened 2017-01-19
Revert "Make tag fields lowercase in atom"

tomrf1:
Found a better way in concierge - https://github.com/guardian/content-api/pull/1837


LATaylor-guardian:
👍 (once build passes)

PR #337, opened 2017-01-19
Adds recipe indexer

tsop14:
- moves indexing logic to common to be shared with reindex sbt project
- removes unused dependencies
- adds indexer sbt project
- adds bash script to run indexer


common/src/main/scala/com/gu/recipeasy/db/DB.scala
```diff
@@ -141,6 +141,11 @@ class DB(contextWrapper: ContextWrapper) {
     contextWrapper.dbContext.run(action)
   }
 
+  val curatedStatuses = List(RecipeStatusVerified, RecipeStatusCurated, RecipeStatusFinalised, RecipeStatusPendingCuration, RecipeStatusPendingVerification, RecipeStatusPendingFinalisation).map(_.name)
```
tomrf1: This list surprises me. Why do we consider pending recipes ready for publication?

shtukas: Hi @tomrf1 

Now that you mention it, you are *perfectly* right that `RecipeStatusPendingCuration` should be excluded from the list. Because in this state that have never been submitted and are simply coming from `Ready` status. This is my mistake, I should have thought about that. 

That having been said, I don't see any problem with `RecipeStatusPendingVerification` and `RecipeStatusPendingFinalisation`.

tomrf1:
Besides the issue with the status list discussed above, this looks fine to me
👍

tomrf1:
I've pushed a commit to remove that state. Are we ready to merge now?
I'd like to get reindexing!

shtukas:
:+1: Thank you!

PR #339, opened 2017-01-30
adds logic to add images for articles with recipes and without images

tsop14:
When ETL is run, after new articles are indexed...

- gets list of articles with recipes saved [A]
- gets list of articles with images saved [B]
- works out articles with missing images by diffing [A] and [B]

### DIFF
--- a/etl/src/main/scala/etl/ETL.scala
+++ b/etl/src/main/scala/etl/ETL.scala
@@ -71,12 +71,11 @@ object ETL extends App {
 
       //add images for articles with recipes and no images
       def articlesWithoutImages: Set[String] = db.getArticlesWithRecipes.toSet diff db.getArticlesWithSavedImages.toSet
-
-      articlesWithoutImages.foreach(a => {
-        capiClient.getResponse(ItemQuery(a)).foreach { itemResponse =>
+      articlesWithoutImages.foreach{ articleId =>
+        capiClient.getResponse(ItemQuery(articleId)).foreach { itemResponse =>
           db.insertImages(ImageExtraction.getImages(itemResponse.content).toList)
         }
-      })
+      }
 
     } finally {
       capiClient.shutdown()
--- a/etl/src/main/scala/etl/ETL.scala
+++ b/etl/src/main/scala/etl/ETL.scala
@@ -68,6 +68,16 @@ object ETL extends App {
       println(s"Finished! End result: $endResult")
       println("Articles with no recipes:")
       endResult.articlesWithNoRecipes.foreach(id => println(s"- https://www.theguardian.com/$id"))
+
+      //add images for articles with recipes and no images
+      def articlesWithoutImages: Set[String] = db.getArticlesWithRecipes.toSet diff db.getArticlesWithSavedImages.toSet
+
+      articlesWithoutImages.foreach(a => {
+        capiClient.getResponse(ItemQuery(a)).foreach { itemResponse =>
+          db.insertImages(ImageExtraction.getImages(itemResponse.content).toList)
+        }
+      })
+
     } finally {
       capiClient.shutdown()
     }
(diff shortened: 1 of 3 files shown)


shtukas:
LGTM, thanks ! :+1:

etl/src/main/scala/etl/ETL.scala
```diff
@@ -68,6 +68,16 @@ object ETL extends App {
       println(s"Finished! End result: $endResult")
       println("Articles with no recipes:")
       endResult.articlesWithNoRecipes.foreach(id => println(s"- https://www.theguardian.com/$id"))
+
+      //add images for articles with recipes and no images
+      def articlesWithoutImages: Set[String] = db.getArticlesWithRecipes.toSet diff db.getArticlesWithSavedImages.toSet
+
+      articlesWithoutImages.foreach(a => {
```
LATaylor-guardian:
```  
    articlesWithoutImages.foreach { articleId =>
         capiClient.getResponse(ItemQuery(articleId)).foreach { itemResponse =>
           db.insertImages(ImageExtraction.getImages(itemResponse.content).toList)
         }
       }
```

etl/src/main/scala/etl/ETL.scala
```diff
@@ -132,7 +142,7 @@ object ETL extends App {
   def getRecipes(content: Content)(insertImages: List[ImageDB] => Unit): Seq[Recipe] = {
 
     val rawRecipes: Seq[RawRecipe] = RecipeExtraction.findRecipes(content.webTitle, content.fields.flatMap(_.body).getOrElse(""))
-    if (rawRecipes.nonEmpty) { insertImages(ImageExtraction.getImages(content, content.id).toList) }
+    if (rawRecipes.nonEmpty) { insertImages(ImageExtraction.getImages(Some(content)).toList) }
```
LATaylor-guardian: Why does this need to be a Some? We already have a Content object.

etl/src/main/scala/etl/ImageExtraction.scala
```diff
@@ -5,8 +5,11 @@ import com.gu.recipeasy.models.{ ImageDB }
 
 object ImageExtraction {
 
-  def getImages(content: Content, articleId: String): Seq[ImageDB] = {
-    findImageElements(content).flatMap(i => extractImageData(i, articleId))
+  def getImages(maybeContent: Option[Content]): Seq[ImageDB] = {
```
LATaylor-guardian: Do you need an option as an argument here?

shtukas:
@tsop14 
I have discussed with @LATaylor-guardian about his suggestions.

What happened is that in `db.insertImages(ImageExtraction.getImages(itemResponse.content).toList)` turns out that `itemResponse.content` is an option. So I was lazy and changed the signature of `getImages` to take an option, but then had to change how `getImages` is called in `insertImages(ImageExtraction.getImages(Some(content)).toList)`

etl/src/main/scala/etl/ETL.scala
```diff
@@ -68,6 +68,16 @@ object ETL extends App {
       println(s"Finished! End result: $endResult")
       println("Articles with no recipes:")
       endResult.articlesWithNoRecipes.foreach(id => println(s"- https://www.theguardian.com/$id"))
+
+      //add images for articles with recipes and no images
+      def articlesWithoutImages: Set[String] = db.getArticlesWithRecipes.toSet diff db.getArticlesWithSavedImages.toSet
+
+      articlesWithoutImages.foreach( articleId => {
```
LATaylor-guardian: When the foreach is over multiple lines it's 
```
.foreach { myVal => 
    ...
}
```

LATaylor-guardian: Normal brackets for oneliner and you don't need thd curly brace after the `=>`

LATaylor-guardian:
👍

PR #342, opened 2017-01-31
fixes processing articles with images missing

tsop14:
Before the capi client was shut down as soon the request returned a promise - this was before any images could be extracted. ☹️
 
* Groups articles missing images into sets of 10
* Requests 10 articles at a time using a string of their ids as a parameter
* Shuts down capi client after Future completed

NB An article might be legitimately missing images e.g. https://www.theguardian.com/lifeandstyle/2007/nov/24/foodanddrink.baking

### DIFF
--- a/etl/src/main/scala/etl/ETL.scala
+++ b/etl/src/main/scala/etl/ETL.scala
@@ -7,7 +7,7 @@ import com.gu.recipeasy.models._
 import com.gu.recipeasy.db._
 
 import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.Await
+import scala.concurrent.{ Await, Future }
 import scala.concurrent.duration._
 import scala.util.{ Failure, Success, Try }
 import cats.kernel.Monoid
@@ -36,6 +36,25 @@ object Progress {
       x.recipesFound + y.recipesFound,
       x.articlesWithNoRecipes ++ y.articlesWithNoRecipes
     )
+  }
+}
+
+object Images {
+
+  def groupArticles(articlesWithoutImages: List[String]): List[String] = {
+    val groups = articlesWithoutImages.grouped(10).toList
+    groups.map(group => group.mkString(","))
+  }
+
+  def getArticleResponse(articleIds: String)(implicit capiClient: GuardianContentClient): Future[SearchResponse] = {
+    capiClient.getResponse(capiClient.search.ids(articleIds).showElements("image"))
+  }
+
+  def getArticlesMissingImages()(implicit db: DB, capiClient: GuardianContentClient): Future[List[SearchResponse]] = {
+    val articlesMissingImages = (db.getArticlesWithRecipes().toSet diff db.getArticlesWithSavedImages().toSet).toList
+    val articleIdsParamGroups: List[String] = groupArticles(articlesMissingImages)
+
+    Future.sequence(articleIdsParamGroups.map(group => getArticleResponse(group)))
   }
 }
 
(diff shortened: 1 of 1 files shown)


shtukas:
Thanks :+1:

etl/src/main/scala/etl/ETL.scala
```diff
@@ -69,16 +88,17 @@ object ETL extends App {
...
-          db.insertImages(ImageExtraction.getImages(itemResponse.content).toList)
+    } finally {
+      val fArticles: Future[List[SearchResponse]] = Images.getArticlesMissingImages()
+      fArticles.map { articles =>
+        articles.foreach { article =>
+          val images = for (results <- article.results) yield ImageExtraction.getImages(Some(results))
+          images.foreach (image => db.insertImages(image.toList))
         }
+      } onComplete {
+        case Success(posts) => capiClient.shutdown()
```
LATaylor-guardian: If you're shutting down regardless you can just do:

```
onComplete {
  case _=> capiClient.shutdown()
}
```

etl/src/main/scala/etl/ETL.scala
```diff
@@ -69,16 +88,17 @@ object ETL extends App {
...
       endResult.articlesWithNoRecipes.foreach(id => println(s"- https://www.theguardian.com/$id"))
 
-      //add images for articles with recipes and no images
-      def articlesWithoutImages: Set[String] = db.getArticlesWithRecipes.toSet diff db.getArticlesWithSavedImages.toSet
-      articlesWithoutImages.foreach { articleId =>
-        capiClient.getResponse(ItemQuery(articleId)).foreach { itemResponse =>
-          db.insertImages(ImageExtraction.getImages(itemResponse.content).toList)
+    } finally {
+      val fArticles: Future[List[SearchResponse]] = Images.getArticlesMissingImages()
+      fArticles.map { articles =>
```
LATaylor-guardian: Should this be: `fArticles.foreach { articles =>`?

etl/src/main/scala/etl/ETL.scala
```diff
@@ -39,6 +39,25 @@ object Progress {
   }
 }
 
+object Images {
+
+  def groupArticles(articlesWithoutImages: List[String]): List[String] = {
```
LATaylor-guardian: This can be private.

etl/src/main/scala/etl/ETL.scala
```diff
@@ -39,6 +39,25 @@ object Progress {
...
+    val groups = articlesWithoutImages.grouped(10).toList
+    groups.map(group => group.mkString(","))
+  }
+
+  def getArticleResponse(articleIds: String)(implicit capiClient: GuardianContentClient): Future[SearchResponse] = {
```
LATaylor-guardian: And this.

etl/src/main/scala/etl/ETL.scala
```diff
@@ -69,16 +88,17 @@ object ETL extends App {
       println("Articles with no recipes:")
       endResult.articlesWithNoRecipes.foreach(id => println(s"- https://www.theguardian.com/$id"))
 
-      //add images for articles with recipes and no images
-      def articlesWithoutImages: Set[String] = db.getArticlesWithRecipes.toSet diff db.getArticlesWithSavedImages.toSet
-      articlesWithoutImages.foreach { articleId =>
-        capiClient.getResponse(ItemQuery(articleId)).foreach { itemResponse =>
-          db.insertImages(ImageExtraction.getImages(itemResponse.content).toList)
+    } finally {
+      val fArticles: Future[List[SearchResponse]] = Images.getArticlesMissingImages()
```
LATaylor-guardian: It feels odd to do this work in a `finally` block. Have you already spoken to @mchv about this given he's in the office?

etl/src/main/scala/etl/ETL.scala
tsop14: IntelliJ suggests you cannot use`on complete` with a foreach...

etl/src/main/scala/etl/ETL.scala
```diff
@@ -39,6 +39,32 @@ object Progress {
...
+  }
+
+  def getArticlesMissingImages()(implicit db: DB, capiClient: GuardianContentClient): Future[List[SearchResponse]] = {
+    val articlesMissingImages = (db.getArticlesWithRecipes().toSet diff db.getArticlesWithSavedImages().toSet).toList
+    val articleIdsParamGroups: List[String] = groupArticles(articlesMissingImages)
+
+    Future.sequence(articleIdsParamGroups.map(group => getArticleResponse(group)))
+  }
+
+  def processArticles(articles: List[SearchResponse])(implicit db: DB) {
```
LATaylor-guardian: Can this have a return type please given it's a public function.

LATaylor-guardian: And maybe call it` insertImages`?

tsop14: It only inserts images if it can find them for the given article. Hopefully most of the time when this runs it will be because the article doesn't have images... so I've gone for `insertImagesFromArticles`

shtukas:
Cool thanks :)

